PR #401, opened 2026-05-14
feat(scheduler): fold scheduler into assistant container, drop HTTP API (#385)

### LINKED ISSUE #385: Fold scheduler into the assistant container; drop the HTTP API
Part of #382.

## Summary

Run the scheduler as a co-process inside the assistant container — no separate image, no exposed port. Admin loses its HTTP client and gains direct write access to the scheduler's automation directory through the existing `${OP_HOME}/config` rw mount; manual triggers and execution logs become file-based.

## Source docs

- Plan section: [`simplify-via-akm.md` § 2 *Fold the scheduler into the assistant container; drop its HTTP API*, lines 28–37](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/simplify-via-akm.md#L28-L37)
- Gap row: [`simplify-via-akm.md` "Scheduler triggers from outside"](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/simplify-via-akm.md#L83-L83)

## Scope

- Delete `core/scheduler/Dockerfile` and the `scheduler` service block from `.openpalm/stack/core.compose.yml`.
- Add a small supervisor entrypoint to the assistant container (`core/assistant/entrypoint.sh` or a Bun supervisor script) that runs OpenCode on `:4096` plus the scheduler subprocess (no port).
- Strip the HTTP layer from `packages/scheduler/src/server.ts`. Keep the croner loop, YAML loader, file-watcher, and library exports (`getSchedulerStatus`, `getLoadedAutomations`, `getExecutionLog`, `triggerAutomation`).
- Add a sentinel-file watcher for manual triggers under `${OP_HOME}/data/scheduler/triggers/<automation>.run`.
- Replace the scheduler HTTP client in `packages/admin/` with filesystem ops: write/edit YAML in `${OP_HOME}/config/automations/`, drop sentinel files for manual triggers, read logs from `${OP_HOME}/logs/`.
- Drop the scheduler base URL and admin↔scheduler auth token from compose env, admin defaults, and `vault/stack/stack.env` template.
- Rewrite `packages/admin/e2e/scheduler.pw.ts` to drive the file-based control plane.

## Dependencies

None — independent. (Coordinates with #382's other sub-issues for compose-file edits but does not block on them.)

## Acceptance

- `docker compose ps` shows no `scheduler` service.
- Inside the assistant container, only `:4096` is listening; the scheduler subprocess runs without binding any port.
- Writing `${OP_HOME}/config/automations/test.yml` from admin causes the scheduler to load it within a couple of seconds (visible in `${OP_HOME}/logs/scheduler.log`).
- Dropping `${OP_HOME}/data/scheduler/triggers/test.run` fires the automation once and removes the sentinel.
- `bun test` in `packages/scheduler` and `packages/admin` is green.

itlackey:
## Summary

Fold the scheduler into the assistant container as a Bun co-process; drop the standalone Docker service and HTTP API entirely.

- **No separate `scheduler` service.** `core/scheduler/Dockerfile` deleted; assistant Dockerfile installs the scheduler source + deps into `/opt/scheduler`. `core/assistant/entrypoint.sh` launches it before `exec opencode` (background process, no port).
- **No HTTP layer.** `packages/scheduler/src/server.ts` now ships only the croner loop, YAML loader, file watcher, library exports — plus a new sentinel-file watcher.
- **Filesystem control plane:**
  - Manual triggers: drop `${OP_HOME}/data/scheduler/triggers/<fileName>.run` → scheduler fires the automation and deletes the sentinel.
  - Logs: `${OP_HOME}/logs/scheduler.log` (entrypoint redirects stdout/stderr).
- **Admin replaces the scheduler HTTP client with filesystem ops:**
  - `POST /admin/automations/:name/run` writes the sentinel.
  - `GET /admin/automations/:name/log` tail-reads `scheduler.log` filtered to the named automation.
- **Compose:** scheduler block removed from `.openpalm/stack/core.compose.yml` and `compose.dev.yml`. Assistant gains `OP_HOME=/openpalm`, `OPENCODE_API_URL=http://localhost:4096`, and three new bind mounts (`config:ro`, `data/scheduler:rw`, `logs:rw`).
- **`OP_OPENCODE_PASSWORD` / scheduler admin token removed from scheduler env wiring** (the co-process uses `OP_ASSISTANT_TOKEN` for `api` actions and reaches OpenCode via loopback with `OPENCODE_AUTH=false`).
- **`CORE_SERVICES` drops `scheduler`** in `packages/lib/src/control-plane/types.ts`.
- **AKM 0.8.0 finding (from issue comments):** `akm task` cannot replace the scheduler in containers — host crontab is invisible inside a container, and there is no `launchd`/`schtasks`. The Bun croner co-process is the only viable path. No change to this PR's scope.

## Test plan

- [x] `bun test` in `packages/scheduler` is green (17 tests; library suite plus sentinel-trigger integration suite that spawns the co-process)
- [x] `bun run test` (all non-admin packages) is green — 849+ tests
- [x] `bun run admin:check` (svelte-check + TypeScript) — 0 errors, 0 warnings
- [x] Admin server vitest suite — 508 tests pass
- [x] `packages/cli/src/install-flow.test.ts` asserts `docker compose config --services` returns `['assistant', 'guardian', 'init', 'memory']` (no scheduler)
- [x] `packages/admin/src/lib/server/lifecycle.vitest.ts` asserts `CORE_SERVICES` has 3 entries and does not contain `scheduler`
- [ ] Stack-dependent Playwright: run `RUN_DOCKER_STACK_TESTS=1 ADMIN_TOKEN=dev-admin-token bun run admin:test:e2e` against a fresh `bun run dev:build` once images rebuild (verifies `docker compose ps` shows no scheduler, sentinel triggers fire, `/admin/automations/:name/log` returns scheduler.log entries)

Closes #385


itlackey:
Addressed review findings in 462b8096:

**Critical**
- Removed all stale `scheduler: 8090` / `scheduler` service refs from `admin-tools/opencode/tools/health-check.ts`, `admin-containers.ts`, `admin-logs.ts`, and `docs/how-it-works.md`. `admin-logs.ts` now notes scheduler logs appear in the assistant container.
- `packages/scheduler/src/server.ts` `processTriggerFile` now rejects filenames containing `/`, `\`, or `..` before any `unlinkSync`, closing the path-traversal hole `path.join` would otherwise allow.
- Added vitest unit tests for both new admin routes (`/admin/automations/:name/run` and `/log`) — 7 + 12 tests covering 401 (unauth), 400 (SAFE_NAME_RE / `..` / `/` / non-numeric/negative/zero limit), 404 (not installed), 202 + sentinel write, empty-log handling, limit application, and bare-name normalization.

**High**
- `core/assistant/entrypoint.sh` now captures the scheduler PID, installs a `SIGTERM`/`SIGINT` trap that forwards termination to the scheduler, and (when the scheduler is running) supervises opencode as a foreground child via `wait` instead of `exec` so the trap is preserved. Falls back to the original `exec` path when the scheduler is not running.
- Changed `executeAssistantAction` default in `packages/lib/src/control-plane/scheduler.ts` from `http://assistant:4096` to `http://localhost:4096` — the docker-network name does not resolve from inside the assistant container. (`OPENCODE_API_URL` is already set explicitly in `core.compose.yml`, but the lib default is the source of truth.)
- Added scheduler co-process integration tests: concurrent-sentinel de-dup (slow shell, two sentinels → one execution), SIGTERM clean-shutdown (exit code 0 within 3s), and hot-reload through the subprocess (drop a new YAML, watcher picks it up, sentinel fires it).

**Medium**
- Renamed `ADMIN_TOKEN` → `ASSISTANT_TOKEN` in `packages/scheduler/src/server.ts` and dropped the stale "admin token" comment.
- Removed dead `getAllExecutionLogs` export (only referenced by its own test); dropped the test and the stale comment in `packages/admin/src/lib/server/scheduler.ts`.

**Low**
- Trimmed the 8-line "we don't use setsid" comment in `entrypoint.sh` to a one-liner explaining the actual WHY (so the new SIGTERM trap can reach the scheduler).
- Left `server.ts` named as-is with a TODO comment — rename to `main.ts` is deferred to a follow-up to keep this PR focused.

**Verification**
- `bun run test` (root) — 851 pass, 0 fail (37 files).
- `bun test packages/scheduler/` — 19 pass, 0 fail (server.test.ts + scheduler.test.ts).
- New admin route tests via `vitest run --project=server` — 19 pass, 0 fail (7 run + 12 log).
- `bun run admin:check` — 0 errors, 0 warnings (650 files).
- `docker compose -f .openpalm/stack/core.compose.yml config --quiet` — lints clean (only unrelated `OP_HOME` defaulting warnings).

(The full `bun run admin:test:unit` script fails on a Playwright chromium binary missing under `~/.cache/ms-playwright/` — verified by `git stash` that this is a pre-existing environment issue, not caused by these changes.)

PR #402, opened 2026-05-14
feat(akm): add periodic akm improve automation (#390)

### LINKED ISSUE #390: Add periodic akm improve automation (replaces akm index --enrich)

## Summary

Schedule `akm index --enrich` on a cadence so akm's enrichment pass — metadata enhancement + memory inference + graph extraction, all shipped in akm-cli 0.7.3 — runs against the shared stash without manual intervention.

## Source docs

- Plan context: [`simplify-via-akm.md` Context — "Crucially: `akm index --enrich` shipped in 0.7.3"](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/simplify-via-akm.md#L9-L9)
- Gap → resolution row: [`simplify-via-akm.md` *LLM fact extraction*](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/simplify-via-akm.md#L70-L70)
- Verification step 7: [`simplify-via-akm.md` lines 124–124](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/simplify-via-akm.md#L124-L124)

## Scope

- Add `${OP_HOME}/config/automations/akm-index.yml` — a small `assistant`-action automation that shells `akm index --enrich`. Cadence configurable via cron expression in the YAML.
- Seed the automation file into the install via `packages/lib/src/control-plane/setup.ts` (idempotent; does not overwrite user edits).
- Document the toggle pathway: users disable enrichment via `akm config set index.infer false` / `index.graph false` rather than removing the automation.

## Dependencies

- Depends on **#386** (akm must be on PATH and the shared stash mounted).
- Coordinates with **#385** (the file-based scheduler must already be running inside the assistant container; not a hard dependency since the automation file format works either way, but verifying this issue is easiest after #385).

## Acceptance

- A clean install seeds `${OP_HOME}/config/automations/akm-index.yml`.
- Trigger the automation via the file-based control plane (sentinel under `${OP_HOME}/data/scheduler/triggers/`); observe `akm index --enrich` runs and `akm events tail` shows the corresponding events.
- A pending memory (no `inferred: true` frontmatter) becomes inferred after a triggered run.
- Errors (e.g., no LLM configured) are logged but do not abort future scheduled runs.

itlackey:
Closes #390

## Summary

- Seed `${OP_HOME}/config/automations/akm-improve.yml` on first install so the shared akm stash gets daily slow-pass LLM maintenance (memory inference, graph extraction, SM-2 reflect/distill cooldowns, schema repair, consolidation) without manual intervention.
- Uses `akm improve` instead of the AKM 0.7.x `akm index --enrich`, which was removed in AKM 0.8.0. Plain `akm index` now runs automatically as a fast post-step of `improve` and on every `akm remember`/import call, so it no longer warrants a separate automation.
- New `seedDefaultAutomations()` helper copies registry catalog files into `config/automations/` only when the destination is missing. Idempotent: user edits survive re-install and upgrade.
- Forwards `AKM_STASH_DIR`, `AKM_DATA_DIR`, `AKM_STATE_DIR`, `AKM_CONFIG_DIR`, and `AKM_CACHE_DIR` through the scheduler's shell-action env allowlist so the automation operates against the same stash root the assistant uses interactively (set declaratively in `core.compose.yml`).

## Operator UX

- Trigger ad-hoc: `touch ${OP_HOME}/data/scheduler/triggers/akm-improve.yml.run` — the scheduler fires the run once and removes the sentinel.
- Disable enrichment without deleting the file: `akm config set index.infer false` / `akm config set index.graph false`.
- Verify a triggered run: `akm events` (state.db) shows `improve_*` events.

## Test plan

- [x] `bun run test` (852 pass / 0 fail) — covers `seedDefaultAutomations` unit test + extended `install-flow` assertion that user-edited automation files survive a re-run of `performSetup`.
- [x] `bun run admin:check` (0 errors, 0 warnings).
- [ ] Manual smoke on the dev stack: drop `akm-improve.yml.run` sentinel, confirm `improve_*` events land in `akm events`.

### DIFF
--- a/.openpalm/registry/automations/akm-improve.yml
+++ b/.openpalm/registry/automations/akm-improve.yml
@@ -1,0 +1,44 @@
+# akm-improve.yml — Periodic AKM stash maintenance pass
+#
+# Runs `akm improve` daily at 03:00 UTC. The `improve` command performs the
+# slow LLM maintenance work that AKM 0.8.0 split out of `akm index`:
+#
+#   - Memory inference (turns pending memories into structured entries)
+#   - Graph extraction (links related notes)
+#   - SM-2 reflect/distill cooldowns
+#   - Schema repair and consolidation
+#
+# Plain `akm index` no longer needs its own automation: it runs as a fast
+# post-step inside `improve`, and on every `akm remember` / import call.
+#
+# Toggle enrichment without removing this file by editing the akm config:
+#   akm config set index.infer false   # disables memory inference
+#   akm config set index.graph false   # disables graph extraction
+#
+# This automation is seeded at install time into ~/.openpalm/config/automations/
+# and is preserved across re-installs (existing user edits are never overwritten).
+#
+# TODO: remove `--auto-accept safe` and the AKM_CONFIG_DIR/config.json seeding
+# step once akm-cli ships `AKM_LLM_PROXY_CMD` (see #345 / akm-cli upstream).
+# When the upstream proxy hook lands, akm can borrow the OpenCode host's
+# existing provider connection and this automation will need no LLM wiring.
+
+name: akm-improve
+description: Run akm improve to consolidate memories, run inference, and update graph extraction
+schedule: "0 3 * * *"
+enabled: true
+
+action:
+  type: shell
+  command:
+    - akm
+    - improve
+    - --auto-accept
+    - safe
+    - --timeout-ms
+    - "3600000"
+  # 1h cap; akm's default is 2h but unattended runs should fail fast and retry
+  # tomorrow rather than block the scheduler for hours on a hung LLM call.
+  timeout: 3600000
+
+on_failure: log
--- a/.openpalm/registry/automations/akm-improve.yml
+++ b/.openpalm/registry/automations/akm-improve.yml
@@ -1,7 +1,9 @@
 # akm-improve.yml — Periodic AKM stash maintenance pass
 #
-# Runs `akm improve` daily at 03:00 UTC. The `improve` command performs the
-# slow LLM maintenance work that AKM 0.8.0 split out of `akm index`:
+# `schedule` runs once during the off-peak window so the slow LLM pass finishes
+# overnight on the previous day's stash activity and stays out of the user's
+# interactive sessions. The `improve` command performs the slow LLM maintenance
+# work that AKM 0.8.0 split out of `akm index`:
 #
 #   - Memory inference (turns pending memories into structured entries)
 #   - Graph extraction (links related notes)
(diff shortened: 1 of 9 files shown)


itlackey:
Addressed reviewer findings in aa2c52f0:

**MEDIUM — Simplify `SEEDED_AUTOMATIONS`** & **Inline `seedDefaultAutomations()`**: Removed both from `registry.ts` and dropped the exports from `packages/lib/src/index.ts`. The seeding (a single `if (!existsSync) writeFile` block for `akm-improve.yml`) now lives at its sole call site in `performSetup()` for the smallest possible surface area.

**HIGH — Scope or remove the try/catch in `performSetup()`**: Removed entirely. The "missing registry catalog" case was already handled inside `getRegistryAutomation()` (returns `null`) and is still gracefully skipped with a warning. The only remaining throws are real filesystem failures (EACCES, ENOSPC, etc.) which should propagate and fail setup loudly, not be silently swallowed.

**MEDIUM — `AKM_*` prefix-match for `SHELL_SAFE_ENV_KEYS`**: Replaced the explicit five-key allowlist with a `SHELL_SAFE_ENV_PREFIXES = ["AKM_"]` prefix check in `executeShellAction()`. Future akm-cli env vars (e.g. `AKM_LLM_PROXY_CMD`) are forwarded automatically with no code changes.

**MEDIUM — Test for env var forwarding**: Added `forwards AKM_* env vars to shell subprocesses` to `packages/scheduler/src/scheduler.test.ts`. The test sets `AKM_STASH_DIR` and a synthetic `AKM_CUSTOM_ARBITRARY` (proves the prefix match, not just the old explicit list), triggers a shell action that writes both values to a file, and asserts the file contents end-to-end.

**MEDIUM — Graceful failure when LLM not configured**: Added `records shell failure (non-zero exit) without crashing the scheduler` to `scheduler.test.ts`. Runs a shell action that exits 1 (simulates `akm improve` failing because no LLM is wired), asserts `triggerAutomation` returns `{ ok: false, error: "shell command failed: ..." }`, the job stays scheduled, and the failure is recorded via `getExecutionLog`. The cron callback already had try/catch around `executeAction`, so no scheduler changes were needed.

**LOW — Redundant cron comment**: Replaced "Runs `akm improve` daily at 03:00 UTC" with a `WHY` rationale ("off-peak overnight window so the slow LLM pass finishes overnight on the previous day's stash activity and stays out of the user's interactive sessions").

**Acceptance**:
- `bun run test` — 853 pass / 0 fail (up from 852: +2 new scheduler tests, -1 deleted seedDefaultAutomations test).
- `bun run admin:check` — 0 errors, 0 warnings, 650 files.
- `OP_HOME=/tmp/op-test docker compose -f .openpalm/stack/core.compose.yml config --quiet` — exit 0.

PR #403, opened 2026-05-14
feat(akm): migrate seeded skills/commands/agents into shared akm stash (#389)

### LINKED ISSUE #389: Migrate seeded skills, commands, and agents into the shared stash

## Summary

Move OpenPalm's bundled skills / commands / agents out of the assistant container image and into the shared akm stash, where they're searchable, versionable, and editable through `akm` without rebuilding the image.

## Source docs

- Plan section: [`simplify-via-akm.md` § 5 *Migrate seeded assets into the stash*, lines 59–62](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/simplify-via-akm.md#L59-L62)
- Configuration ownership rule (background): existing config-doesn't-overwrite-user-edits behavior in `packages/lib/src/control-plane/setup.ts`.

## Scope

- Move assets from `core/assistant/opencode/{commands,agents,skills}/` into stash subdirectories with `.stash.json` metadata under `${OP_HOME}/data/stash/{commands,agents,skills}/`.
- Update `packages/lib/src/control-plane/setup.ts` to seed those into the shared stash on first install (idempotent — does not overwrite user edits).
- Update `core/assistant/opencode/system.md` to reference the assets via akm refs (`agent:<name>`, `command:<name>`, `skill:<name>`) instead of bundled paths.
- Remove the assistant-image bundling from `core/assistant/Dockerfile` for the migrated assets.

## Dependencies

- Depends on **#386** (akm CLI + shared stash must exist before assets can be seeded).

## Acceptance

- `docker compose exec assistant akm list --type {skill,command,agent}` shows the seeded assets.
- A fresh `openpalm install` on a clean directory seeds the stash without overwriting any pre-existing user edits.
- Assistant slash-commands resolve via the akm refs.

itlackey:
Closes #389

## Summary

Move the built-in assistant skills out of the assistant container image and into the shared akm stash so they're searchable, editable, and resolvable via `akm show <ref>` without rebuilding the container.

- Moved `core/assistant/opencode/skills/config-diagnostics.md` to `.openpalm/stash-seeds/skills/config-diagnostics/SKILL.md` with akm frontmatter (`name`, `type`, `description`, `when_to_use`).
- Added `seedStashAssets()` and `STASH_SEED_PATHS` in `@openpalm/lib` — idempotent seeder that copies built-in stash assets into `${OP_HOME}/data/stash/` and **never overwrites user edits**.
- Embedded seeds via Bun text imports in `packages/cli/src/lib/embedded-assets.ts`; `seedEmbeddedAssets()` materializes the stash on every fresh CLI install (works offline).
- Rewrote `core/assistant/opencode/system.md` to reference the seeded skill by akm ref (`skill:config-diagnostics`) instead of a bundled path.
- Added a Dockerfile comment making explicit that skills/commands/agents now live in the bind-mounted stash, not in the image; the `COPY core/assistant/opencode /etc/opencode` line no longer carries any bundled skills (the source directory was removed).

## Migrated assets

Only one seed shipped before this PR (`skills/config-diagnostics.md`). No commands or agents were bundled; the new `.openpalm/stash-seeds/{commands,agents}/` slots are scaffolded for future additions.

## Test plan

- [x] `bun run test` — 859 / 859 pass
- [x] `bun run admin:check` — 0 errors, 0 warnings
- [x] New `packages/lib/src/control-plane/core-assets.test.ts` covers `seedStashAssets()` (write, idempotency, nested paths, empty input) and verifies `STASH_SEED_PATHS` shape
- [x] New `install-flow.test.ts` cases assert (a) every embedded seed lands in `data/stash/...` on first install, and (b) a user-edited seed is preserved across re-install
- [ ] Manual post-merge verification: `docker compose exec assistant akm list --type skill` shows `config-diagnostics`
- [ ] Manual post-merge verification: fresh `openpalm install` on a clean directory seeds the stash; a second run with a user-edited skill preserves the edit

### DIFF
--- a/.openpalm/stash-seeds/README.md
+++ b/.openpalm/stash-seeds/README.md
@@ -1,0 +1,32 @@
+# Stash Seeds
+
+Seed assets for the shared akm stash. These files are copied into
+`${OP_HOME}/data/stash/` on first install by `seedStashAssets()`.
+
+## Layout
+
+```
+stash-seeds/
+├── skills/        # Skills (directories with SKILL.md + frontmatter)
+├── commands/      # Slash commands (flat .md files)
+└── agents/        # Agent personas (flat .md files)
+```
+
+## Conventions
+
+- **Skills** are directories containing a `SKILL.md` with YAML frontmatter
+  (`name`, `type: skill`, `description`, `when_to_use`). Supporting files
+  live alongside `SKILL.md`. Resolved via `akm show skill:<name>`.
+- **Commands** are flat markdown files with YAML frontmatter
+  (`name`, `type: command`, `description`, `when_to_use`).
+  Resolved via `akm show command:<name>`.
+- **Agents** are flat markdown files with YAML frontmatter
+  (`name`, `type: agent`, `description`, `when_to_use`).
+  Resolved via `akm show agent:<name>`.
+
+## Seeding Rules
+
+- First install copies every seed into `${OP_HOME}/data/stash/<type>/<name>...`.
+- **Subsequent installs never overwrite existing files** — user edits win.
+- Seeds are embedded into the CLI binary via Bun text imports, so a fresh
+  install works offline.
--- a/.openpalm/stash-seeds/skills/config-diagnostics/SKILL.md
+++ b/.openpalm/stash-seeds/skills/config-diagnostics/SKILL.md
@@ -1,3 +1,14 @@
+---
+name: config-diagnostics
+type: skill
+description: Diagnose OpenPalm configuration issues (missing API keys, validation errors, connection problems) using the admin /admin/config/validate endpoint and schema metadata, without exposing any actual secret values.
+when_to_use: Use when the user reports configuration issues, missing API keys, validation errors, or connection problems and needs guidance on what to fix without leaking secrets.
+license: Proprietary
+metadata:
+  author: openpalm
+  version: "1.0"
+---
+
 # Config Diagnostics Skill
 
 When a user asks about configuration issues, connection problems, missing API
--- a/core/assistant/Dockerfile
+++ b/core/assistant/Dockerfile
@@ -151,6 +151,11 @@ RUN BUN_INSTALL=/usr/local bun add -g "akm-cli@${AKM_CLI_VERSION}" \
 COPY --from=varlock-fetch /usr/local/bin/varlock /usr/local/bin/varlock
 RUN mkdir -p /usr/local/etc/varlock && mkdir -p /etc/opencode
 COPY .openpalm/vault/redact.env.schema /usr/local/etc/varlock/.env.schema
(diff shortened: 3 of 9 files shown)


itlackey:
## Reviewer findings addressed (206ee502)

### HIGH — Path-traversal guard in `seedStashAssets()`
- Canonicalize the computed target with `path.resolve()` and reject anything whose normalized path is not equal to `data/stash/` itself or a child under `data/stash/${path.sep}`.
- Two new unit tests cover the rejection path: `"../../etc/cron.d/evil"` and `"skills/../../../escape.md"` both throw `Seed path escapes stash dir: ...`.

### MEDIUM — Removed `STASH_SEED_PATHS`
- Deleted the constant from `packages/lib/src/control-plane/core-assets.ts`, its re-export from `packages/lib/src/index.ts`, and the shape-test from `core-assets.test.ts`. The manifest was never enforced at runtime — `refreshCoreAssets()` iterates `MANAGED_ASSETS` only — and `EMBEDDED_STASH_SEEDS` in the CLI already owns the seed list.
- Updated the comment in `embedded-assets.ts` to point at `.openpalm/stash-seeds/` as the source of truth (no constant cross-reference).
- Updated `install-flow.test.ts` to drop the `STASH_SEED_PATHS` import and rely directly on `EMBEDDED_STASH_SEEDS`.

### MEDIUM — Simplify `seedStashAssets()`
Kept the parameter rather than picking strict option (1) or (2). Rationale: removing the parameter would force the seed content to live inside lib (duplicating the `.openpalm/stash-seeds/` source of truth and dragging Bun text imports into a package that admin consumes via Vite). Keeping the parameter is also what makes the new path-traversal guard directly testable inside lib without standing up a fake CLI fixture. The "smaller API surface" goal is otherwise met by removing `STASH_SEED_PATHS`.

### LOW — `MANAGED_ASSETS` comment
Added the two-line comment above `MANAGED_ASSETS` explaining that stash seeds intentionally skip `refreshCoreAssets()` so user edits always win.

### LOW — Read-only directory test
New test `chmod`s the stash dir to `0o555`, calls `seedStashAssets`, and asserts it throws. Skipped automatically when `process.getuid?.() === 0` (CI may run as root).

### LOW — Install-flow content assertion
`seedEmbeddedAssets copies built-in stash skills on first install` now also asserts that the seeded `SKILL.md` contains `name: config-diagnostics` AND `type: skill` in the frontmatter AND a non-empty body after the closing `---`. Proves the install actually ran the seeder end-to-end rather than merely creating an empty file.

### Verification
- `bun run test` — 860 pass / 0 fail (was 833 pass on parent)
- `bun run admin:check` — 0 errors, 0 warnings, 634 files

Issue #386, opened 2026-05-07
Install akm-cli in trusted containers; share stash between admin and assistant; give guardian its own


itlackey:
Part of #382. **Foundational** — blocks the rest of Epic #382's sub-issues.

## Summary

Bake `akm-cli@^0.7.4` into the guardian, assistant, and admin Dockerfiles and bind-mount stash directories from the host. Admin and assistant share `${OP_HOME}/data/stash` (rw both sides) so admin can browse and edit user-facing assets. Guardian gets a separate `${OP_HOME}/data/guardian-stash` (rw, isolated) for operator-only artifacts.

## Source docs

- Plan section: [`simplify-via-akm.md` § 3 *Install akm in the trusted containers; share a stash between admin and assistant; give guardian its own*, lines 39–52](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/simplify-via-akm.md#L39-L52)
- Context: [`simplify-via-akm.md` Context, lines 7–9](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/simplify-via-akm.md#L7-L9) — akm 0.7.x feature surface and the fact that `akm-opencode` no longer auto-installs the CLI on load.

## Scope

- `core/guardian/Dockerfile`, `core/assistant/Dockerfile`, `core/admin/Dockerfile`: add a pinned `bun add -g akm-cli@^0.7.4` (or equivalent) install step.
- `.openpalm/stack/core.compose.yml`: extend the existing assistant stash mount; add `${OP_HOME}/data/stash → /akm` rw on admin; add `${OP_HOME}/data/guardian-stash → /akm` rw on guardian; set `AKM_STASH_DIR` consistently per container.
- `packages/lib/src/control-plane/setup.ts`: create `${OP_HOME}/data/guardian-stash/` on first install, with `OP_UID:OP_GID` ownership. Seed guardian's HMAC channel secrets into an `akm vault` inside the guardian stash (operator-only).
- Document in `docs/technical/environment-and-mounts.md` (rewrite covered separately under #383).

## Dependencies

None — foundational for the rest of Epic #382.

## Blocks

- Replace memory service and overlapping plugin tools (next sub-issue)
- Migrate seeded skills/commands/agents into the shared stash
- Migrate vault/user secrets into akm vault
- Add periodic `akm index --enrich` automation

## Acceptance

[...]

PR #404, opened 2026-05-14
feat(akm): mirror vault/user secrets into akm secret store for UI visibility (#388)

### LINKED ISSUE #388: Migrate vault/user secrets into akm vault inside the shared stash

## Summary

Move user-managed secrets from `${OP_HOME}/vault/user/*.env` into an `akm vault` resident in the shared assistant↔admin stash. Operator-only secrets in `${OP_HOME}/vault/stack/` stay where they are — different trust scope.

## Source docs

- Plan section: [`simplify-via-akm.md` § 6 *Move user secrets into `akm vault`; keep stack secrets in plain env files*, lines 63–67](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/simplify-via-akm.md#L63-L67)
- Trust-model justification (companion plan): [`configuration-simplification.md` § Vault layout — collapse, lines 99–104](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/configuration-simplification.md#L99-L104)
- Companion debate: [`configuration-simplification.md` § Tensions resolved — "User vault → akm vault concentrates secrets"](https://github.com/itlackey/openpalm/blob/c2e7b6483bffa82a3e9ca4d467d1189270c05247/docs/plans/configuration-simplification.md#L65-L73)

## Scope

- During install, write existing `${OP_HOME}/vault/user/*.env` entries via `akm vault create/set` into a stash-resident vault. One-time idempotent migration in `packages/lib/src/control-plane/setup.ts`.
- After successful migration, delete the `vault/user/` host directory.
- Replace the `load_vault` assistant tool with calls to `akm_vault load`.
- Remove the `${OP_HOME}/vault/user → /etc/vault` mounts from compose (assistant + apprise file mount).
- Update `.openpalm/vault/README.md` to drop user-vault and varlock references.

## Dependencies

- Depends on **#386** (akm CLI must be available before secrets can be written via `akm vault`).

## Blocks

- **#391** (varlock removal) — depends on this so user-secret hygiene moves entirely to akm.
- **#394** (move owner name/email out of env into akm vault profile) — uses the same vault.

## Acceptance

- `${OP_HOME}/vault/user/` does not exist after a successful install on a new system.
- `akm vault list` from inside assistant or admin shows the migrated entries.
- `akm_vault load` populates env for an assistant tool call.
- Existing channel HMAC secrets in `vault/stack/guardian.env` and admin/assistant tokens in `vault/stack/stack.env` are untouched.

itlackey:
Closes #388

## Summary

Phase 1 of #388 — keep `${OP_HOME}/vault/user/user.env` as the runtime source of truth for Docker Compose `env_file` consumption while **mirroring** its keys into the shared akm secret store at `vault:user`. The assistant and admin UI can now browse/edit user secrets through the same `akm vault` interface they already use for every other shared secret, without breaking the existing compose-mounted bind that production OpenCode containers depend on.

The issue comments captured a key correction (no akm load subcommand exists in akm 0.8.0) — this PR implements the **recommended Option A** described in those comments.

## What changed

- `packages/lib/src/control-plane/akm-vault.ts` *(new)* — wraps the akm CLI (create / set / path subcommands) with an idempotent key-by-key diff. Skips cleanly when akm is missing.
- `performSetup()` and `applyUpgrade()` invoke `mirrorUserVaultToAkm(state)` (best-effort, never blocks install/upgrade). The upgrade hook covers the 0.10.x migration acceptance criterion.
- `packages/admin/src/routes/admin/secrets/user-vault/+server.ts` *(new)* — `GET/POST/DELETE` for `vault:user`. Lists keys only (values are never returned). Writes are mirrored into the akm store **and** `user.env` so the compose runtime stays in sync.
- `load_vault` assistant tool now resolves the vault path through the akm CLI's `path` subcommand, falling back to `/etc/vault/user.env` when akm is unavailable.
- `state.setupToken` is held in memory only (#388 sec B.2). `${dataDir}/setup-token.txt` is no longer written.
- Explicit non-change comment for `OP_OPENCODE_PASSWORD` compose plumbing (#388 sec B.1) so a reviewer doesn't strip it as part of vault cleanup.

## Non-changes (intentional)

- `vault/stack/stack.env` and `vault/stack/guardian.env` are **not** mirrored. Migrating them would break guardian's HMAC `env_file` hot-reload contract.
- The `${OP_HOME}/vault/user -> /etc/vault` compose mount is **kept**. Removing it requires entrypoint sourcing changes and is deferred to Phase 2.

## Test plan

- [x] `bun run test` -- 856 pass / 0 fail (from repo root, lib + cli + guardian + sdk + channels)
- [x] `bun run admin:check` -- 0 errors / 0 warnings (637 files)
- [x] Admin vitest `server` project -- 534 pass (40 files), includes the 7 new user-vault route tests
- [x] New `akm-vault.test.ts` -- 5 pass (idempotency, change-only writes, missing/empty user.env, ref string)
- [x] `docker compose -f .openpalm/stack/core.compose.yml config --quiet` -- clean (exit 0)
- [x] Migration: seeded a fake 0.10.x layout with `GROQ_API_KEY=xyz-test-value-1`, ran `mirrorUserVaultToAkm()`, verified the key landed in the akm vault file AND `user.env` was preserved. Reran -- all keys reported as `unchanged`.
- [x] `state.setupToken` is no longer written to disk (verified no production callers read `data/setup-token.txt`)

## Phase 2 follow-up (deferred -- not this PR)

- Retire the `${OP_HOME}/vault/user -> /etc/vault` compose mount
- Source the akm vault path from an entrypoint instead of bind-mounting
- Delete `${OP_HOME}/vault/user/` once Phase 1 has shipped and stabilized

Generated with Claude Code (claude.com/claude-code)

### DIFF
--- a/.openpalm/stack/core.compose.yml
+++ b/.openpalm/stack/core.compose.yml
@@ -99,6 +99,9 @@ services:
       # reaches OpenCode via http://localhost:4096.
       # If you change OP_ASSISTANT_BIND_ADDRESS to 0.0.0.0, you MUST set
       # OP_OPENCODE_PASSWORD in stack.env and set OPENCODE_AUTH to "true".
+      # NON-CHANGE (#388 §B.1): OP_OPENCODE_PASSWORD compose plumbing
+      # is intentional for the LAN-exposed assistant case. Do NOT strip
+      # it as part of vault cleanup.
       OPENCODE_AUTH: "false"
       OPENCODE_ENABLE_SSH: ${OPENCODE_ENABLE_SSH:-0}
       TERM: xterm-256color
--- a/core/assistant/opencode/system.md
+++ b/core/assistant/opencode/system.md
@@ -21,6 +21,6 @@ For information about managing OpenPalm view @openpalm.md
 
 ## Secrets & Environment
 
-- Use `load_vault` to load user secrets from `/etc/vault/user.env` — this is the primary tool for accessing API keys, owner info, and other user-configured secrets.
+- Use `load_vault` to load user secrets (prefers the shared akm `vault:user` store, falls back to `/etc/vault/user.env`) — this is the primary tool for accessing API keys, owner info, and other user-configured secrets.
 - Use `load_env` only for ad-hoc `.env` files in the `/work` directory (workspace). It cannot read files outside `/work`.
 - Never display, log, or store secret values.
--- a/packages/admin/src/routes/admin/secrets/user-vault/+server.ts
+++ b/packages/admin/src/routes/admin/secrets/user-vault/+server.ts
@@ -1,0 +1,210 @@
+/**
+ * /admin/secrets/user-vault — read/write the shared akm user vault.
+ *
+ * Phase 1 of #388: this endpoint surfaces the same `vault:user` keys
+ * that `mirrorUserVaultToAkm()` populates during install/upgrade. The
+ * underlying compose runtime source of truth (`vault/user/user.env`)
+ * is kept in sync on writes so Docker Compose env_file resolution
+ * never diverges from what the admin UI shows.
+ *
+ * NOTE: We deliberately do NOT route admin secret writes through this
+ * endpoint by default — operator-managed `stack.env` secrets remain in
+ * the existing `/admin/secrets` plaintext/pass backends. This endpoint
+ * is scoped to user-extension keys only.
+ */
+import type { RequestHandler } from './$types';
+import { getState } from '$lib/server/state.js';
+import {
+  errorResponse,
+  getActor,
+  getCallerType,
+  getRequestId,
+  jsonResponse,
+  parseJsonBody,
+  jsonBodyError,
+  requireAdmin,
+} from '$lib/server/helpers.js';
+import {
+  appendAudit,
+  ensureAkmUserVault,
+  mirrorUserVaultToAkm,
+  readAkmUserVaultFile,
+  parseEnvFile,
+  mergeEnvContent,
+} from '@openpalm/lib';
+import { existsSync, readFileSync, writeFileSync, mkdirSync } from 'node:fs';
+import { execFile as execFileCb } from 'node:child_process';
+import { promisify } from 'node:util';
+
+const execFile = promisify(execFileCb);
+
+const KEY_RE = /^[A-Za-z_][A-Za-z0-9_]*$/;
+
+function buildAkmEnv(state: ReturnType<typeof getState>): NodeJS.ProcessEnv {
+  const stashRoot = `${state.dataDir}/stash`;
+  return {
+    ...process.env,
+    AKM_STASH_DIR: stashRoot,
+    AKM_DATA_DIR: `${stashRoot}/.data`,
+    AKM_STATE_DIR: `${stashRoot}/.state`,
+    AKM_CONFIG_DIR: `${stashRoot}/.config`,
+    AKM_CACHE_DIR: `${state.dataDir}/akm-cache`,
+  };
+}
+
+function writeUserEnvKey(vaultDir: string, key: string, value: string): void {
+  const userEnvPath = `${vaultDir}/user/user.env`;
+  mkdirSync(`${vaultDir}/user`, { recursive: true, mode: 0o700 });
+  const existing = existsSync(userEnvPath) ? readFileSync(userEnvPath, 'utf-8') : '';
+  const merged = mergeEnvContent(existing, { [key]: value });
+  writeFileSync(userEnvPath, merged.endsWith('\n') ? merged : merged + '\n', { mode: 0o600 });
+}
+
+/**
+ * GET — list keys in the akm vault:user store. Values are NEVER returned
+ * so this endpoint behaves identically whether the underlying backend
+ * exposes plaintext or encrypted secrets.
+ */
+export const GET: RequestHandler = async (event) => {
+  const requestId = getRequestId(event);
+  const authError = requireAdmin(event, requestId);
+  if (authError) return authError;
+
+  const state = getState();
+  const actor = getActor(event);
+  const callerType = getCallerType(event);
+
+  // Ensure mirror is up to date before listing so a freshly-installed
+  // stack shows the seeded keys without a separate /upgrade call.
+  try { await mirrorUserVaultToAkm(state); } catch { /* best-effort */ }
+
+  const vaultPath = await ensureAkmUserVault(state);
+  const userEnvPath = `${state.vaultDir}/user/user.env`;
+  const akmKeys = vaultPath ? Object.keys(readAkmUserVaultFile(vaultPath)) : [];
+  const fileKeys = Object.keys(parseEnvFile(userEnvPath));
+  const merged = Array.from(new Set([...akmKeys, ...fileKeys])).sort();
+
+  appendAudit(
+    state,
+    actor,
+    'secrets.user-vault.list',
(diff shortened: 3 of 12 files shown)


itlackey:
## Reviewer findings addressed in `fe72f8ad`

### CRITICAL — secret value leak via `/proc/<pid>/cmdline`

`akm 0.8.x` accepts vault-write values only on argv. Replaced the
`execFile("akm", ["vault", "set", ref, key, value])` calls in both
`packages/lib/src/control-plane/akm-vault.ts` and the admin route with
**direct .env file writes** through two new lib helpers:

- `writeAkmVaultKey(state, key, value)` — resolves the vault path via
  `vault path` (no secret on argv), then writes via `mergeEnvContent`
  + `writeFileSync`.
- `deleteAkmVaultKey(state, key)` — symmetric remove, using the new
  `removeEnvKey` helper.

The vault file format is unchanged (plain `.env`), so `vault list` /
`vault path` keep working. Added a regression test that mocks
`execFile`, runs `mirrorUserVaultToAkm` with a known sentinel value,
and asserts **no execFile call contains the value on argv**.

### HIGH — duplicate `buildAkmEnv` in admin route

Exported `buildAkmEnv` from `@openpalm/lib`. Removed the verbatim
duplicate from `packages/admin/src/routes/admin/secrets/user-vault/+server.ts`.
Per `CLAUDE.md` control-plane logic must live in `packages/lib/`.

### HIGH — atomicity / silent catch on POST and DELETE

The admin route's mirror failures previously set `mirrored: false` and
returned `ok: true` with no operator signal. Now:

- The catch logs `warn` with the key name and a reason field (never
  the value).
- The response includes an `error` field describing the failure type
  (`akm_unavailable` or the caught error message).
- `mirrored: false` is still returned — the user.env write succeeded
  and is the runtime source of truth, but callers can see the
  divergence and decide whether to retry.

### HIGH — dead `resolveAkmUserVaultPath` export

Removed. The lone candidate caller (`load_vault.ts` in `assistant-tools`)
doesn't depend on `@openpalm/lib`, so the export served no consumer.
The inline `resolveVaultPath` in `load_vault.ts` was updated to match
the lib version's `existsSync` guard so a stale ref cleanly falls
through to the legacy mount.

### HIGH — `readAkmUserVaultFile` "tests/diagnostics only" comment

Removed the misleading comment. The admin GET handler is a real
consumer, so the function is intentional public API.

### HIGH — 0.10.x upgrade-path test

Added `migrates a fake 0.10.x layout idempotently (upgrade-path contract)`
in `akm-vault.test.ts`. Seeds a pre-0.11 `vault/user/user.env`, calls
the same `mirrorUserVaultToAkm` that `applyUpgrade` invokes, and
asserts:

1. Keys are written to the akm vault file
2. `user.env` is unchanged (Phase 1 keeps both in sync)
3. Second run reports every key as `unchanged` (idempotency)

This is the operation `applyUpgrade()` performs after
`refreshCoreAssets`+`reconcileCore`. Testing `applyUpgrade` directly
would require Docker; the mirror path is tested in isolation here and
the lifecycle wiring (`await mirrorUserVaultToAkm(state)`) is
unchanged on this branch.

### MEDIUM — `load_vault` exposed absolute path to LLM

Tool results now use a symbolic `source` label (`akm:vault:user` or
`fallback:/etc/vault`) instead of the full filesystem path. The LLM
never needs the stash absolute path.

### MEDIUM — mirror called on every GET

Removed `mirrorUserVaultToAkm(state)` from the GET handler. Mirror is
a lifecycle operation (install/upgrade); GET is read-only.

### MEDIUM — DELETE left empty-valued key in user.env

DELETE now fully removes the line from `user.env` via the new
`removeEnvKey` helper. GET no longer needs to filter empty-valued
keys (but defensively still does — cheap and avoids future drift).
Added a `DELETE followed by GET no longer lists the key` test.

### MEDIUM — `AKM_CACHE_DIR` placement

Added a one-line WHY comment on `buildAkmEnv`: `AKM_CACHE_DIR` lives
outside `stash/` because cache is regenerable and shouldn't be
indexed alongside real stash assets (matches the compose mount
introduced by #386).

### LOW — hardcoded `'vault:user'`

Admin route now imports `AKM_USER_VAULT_REF` from lib and uses it
everywhere.

## CI coverage note

The `it.skipIf(!AKM_AVAILABLE)` gated tests skip silently when the
akm CLI is not on PATH. CI does not install akm today, so the
direct-write branches and the argv-leak regression test are exercised
only by local developers. **Follow-up:** either install akm in CI or
provide an in-memory mock — to be tracked in a separate issue.

## Upstream issue filed

https://github.com/itlackey/akm/issues/361 — proposes `--stdin` and
`--from-env` flags so future versions of the CLI can accept vault
values without argv exposure. If accepted, we can switch the mirror
back to a CLI-only path with no value crossing argv (the .env file
format is byte-compatible either way).

## Test summary

- `bun test` (packages/lib): **408 pass, 0 fail** (4 new akm-vault
  tests, 4 new env tests for `removeEnvKey`)
- `vitest run --project server` (packages/admin): **535 pass, 0 fail**
  (updated DELETE semantics; new GET-after-DELETE assertion)
- `bun run admin:check`: **0 errors, 0 warnings, 638 files**
- `bun test` (packages/assistant-tools): **17 pass, 0 fail**
- `bun run test` from repo root: 862 pass / 1 fail. The 1 failure
  (`cli main > backs up the current OP_HOME before install --force
  rewrites assets`) is **pre-existing** on `c77efb8f` (confirmed by
  stashing the changes and re-running) — unrelated to this PR.

itlackey:
### CI hang fix — `cli main > backs up the current OP_HOME before install --force rewrites assets`

**Root cause** (commit `fab9234d`):

`mirrorUserVaultToAkm` awaited promisified `child_process.execFile` calls with no wall-clock bound. In the Bun runtime, `child_process.execFile` is implemented on top of `Bun.spawn`. The CLI install tests stub `Bun.spawn` (`packages/cli/src/main.test.ts` `mockDockerCli`) and return a fake child whose `stdout`/`exited` never resolve — so the promisified `execFile("akm", ["--version"])` inside `akmAvailable()` hangs forever. The same failure mode would hit production if `akm` itself ever wedged.

The non-`--force` install tests masked this because they leave `user.env` empty, so `mirrorUserVaultToAkm` returns early at the "user.env empty" check before reaching `akmAvailable`. The `--force` test seeds `EXISTING=1` in `user.env`, which is the first code path that actually shells out.

**Fix**: every akm invocation now goes through `execAkm()`, which wraps `execFile` in a `Promise.race` against an unref'd `setTimeout(2000)`. Real akm completes in ~150ms; a stuck binary fast-rejects, `akmAvailable` swallows it as "akm not on PATH", and the mirror returns a clean skip result. Security contract (no secret values on argv) is unchanged — the `execAkm` helper accepts the same `args` array, never strings.

`execFile`'s built-in `timeout` option is insufficient because it signals the child via the spawn handle, which the stub never wires up; the wall-clock `Promise.race` is the only reliable bound for stubbed runtimes.

**Verification**:
- `bun test packages/cli/src/main.test.ts` — 28/28 pass
- `bun run test` — 864/864 pass (including a new regression test in `akm-vault.test.ts` that stubs `Bun.spawn` the same way the CLI install tests do and asserts the mirror returns in <4s with a skip result)
- `bun run admin:check` — 0 errors / 0 warnings
- `bun run admin:test:unit` — 541/541 pass

Issue #409, opened 2026-05-14
Over-engineering audit: 72 findings from 6-reviewer end-to-end review


itlackey:
## Summary

A 6-agent critical review pass surfaced **~72 findings** of unjustified complexity, dead code, real bugs, and stale documentation across the OpenPalm codebase. Each reviewer focused on a distinct surface and looked for things that could be refactored away **without losing functionality**.

### B1 — `currentValueForTarget` has identical ternary branches
**File**: `packages/lib/src/control-plane/secret-backend.ts:86-92`
Both branches of the `target.scope === 'system' ? ... : ...` ternary call `readStackEnv(state.vaultDir)`. The user-scope branch was meant to read `user.env` but reads `stack.env` instead. `currentValueForTarget` is broken for user-scope secrets and silently returns the wrong file's value.

### B2 — Admin logger duplicates lib's `createLogger` WITHOUT redaction
**File**: `packages/admin/src/lib/server/logger.ts:3`
Lib's `createLogger` applies `redactExtra()` to the `extra` payload. Admin's copy at `lib/server/logger.ts` skips that. **Secrets can leak through admin route logs.** Delete the admin copy; import from `@openpalm/lib`.

### B3 — `init` chmod ordering leaves subdirs world-readable
**File**: `.openpalm/stack/core.compose.yml:29`
The `mkdir -p` creates `/data/guardian-stash/.data` (etc.) at default 755, then `chmod 700 /data/guardian-stash`. The subdirs are reachable only through the 700 parent today, but if the parent mode ever changes (overlay), the subdirs are exposed. Either chmod the subdirs explicitly or set umask before mkdir.

### B4 — `admin-config.ts` is dead and misleading
**File**: `packages/admin-tools/opencode/tools/admin-config.ts`
- `get_access_scope` calls `GET /admin/connections/status` (wrong endpoint)
- `set_access_scope` always returns a hard error
Both are registered tools the LLM can call. Delete the file and both registrations.

---

### A1 — Shared base image for admin/assistant/guardian Dockerfiles
**Files**: `core/{admin,assistant,guardian}/Dockerfile`
Bun + akm-cli + OpenCode install steps are duplicated 3 times. **Worse**: `OPENCODE_VERSION` is `1.3.3` in assistant but `1.2.24` in admin (drift!). A shared `openpalm/base` image cuts build time and eliminates skew.

### A2 — `SecretBackend` interface with 2 impls and no polymorphic use
[...]

PR #421, opened 2026-05-15
feat(vault): complete Phase 2 of #388 + migrate to akm secret-store stdin pattern (closes #406)

### LINKED ISSUE #406: Switch akm secret-store writes from direct file I/O to CLI stdin pattern
## Background

PR #404 (issue #388) implemented vault/user → akm secret-store mirroring as Phase 1 of #388. Because the relevant akm 0.8.0-rc1 subcommand accepted values only as positional argv (which leaks via `/proc/<pid>/cmdline`), the implementation bypassed the CLI entirely and writes values directly to the underlying `.env` file using Node `fs` APIs.

Upstream akm `release/0.8.0` has since shipped fixes that resolve this:
- `eff3e37` — added `--stdin` and `--from-env` flags (closes itlackey/akm#361)
- `c50f9f4` — made stdin the default; removed the positional `value` argument AND the `KEY=VALUE` combined form entirely

The new CLI shape (no values cross argv anymore):
- Default: `<ref> <key>` reads value from stdin
- Alternative: `<ref> <key> --from-env VAR` reads from a named env var
- The old positional and combined forms are gone

## Impact

Current PR #404 implementation works correctly. This issue is a forward-looking cleanup, not a bug fix.

## Why migrate

The current direct-file-write workaround in `packages/lib/src/control-plane/akm-vault.ts` (`writeAkmVaultKey`, `deleteAkmVaultKey`) is:
- Secure (no argv exposure)
- Functional today

But it couples OpenPalm tightly to akm's internal `.env` file format. If akm later adds encryption, metadata, or any wrapping, our direct writes break silently.

Switching to the CLI stdin pattern:
- Follows the official, supported API
- Future-proofs against file-format changes
- Removes ~30 lines of file-format-specific code from `akm-vault.ts`

## Scope

1. **Bump the pinned akm-cli version** in `core/{assistant,admin,guardian}/Dockerfile` from `^0.8.0-rc1` to whatever stable version ships `c50f9f4`. (Currently both commits are on `release/0.8.0` HEAD, not yet tagged.)

2. **Replace `writeAkmVaultKey`** in `packages/lib/src/control-plane/akm-vault.ts` with a spawn-based call where the value is written to stdin:
```ts
const child = spawn("akm", ["vault", "<write-subcommand>", AKM_USER_VAULT_REF, key], { env, stdio: ["pipe", "pipe", "pipe"] });
child.stdin.end(value);  // value never appears on argv
await new Promise<void>((resolve, reject) => {
  child.on("exit", (code) => code === 0 ? resolve() : reject(new Error(`akm <write> failed: ${code}`)));
});
```

3. **Replace `deleteAkmVaultKey`** with the akm `unset` subcommand (already argv-safe since no value crosses the boundary).

4. **Delete the `mergeEnvContent` / direct-write code path** once the CLI path is verified working.

5. **Keep the existing `akmAvailable()` 2s timeout** (added during PR #404's hang fix) — still needed for non-interactive test environments where `Bun.spawn` is stubbed.

6. **Update tests**: the regression test in `akm-vault.test.ts` that asserts "never passes secret values via execFile argv" should remain — argv-clean is still the security contract — but now verify values flow via stdin or `--from-env`.

## Acceptance

- `grep -rn 'writeFileSync.*vault' packages/lib/src/control-plane/akm-vault.ts` returns nothing (no direct file writes)
- All mutations go through `spawn("akm", ["vault", ...])` with values via stdin or `--from-env`
- `bun run test` green
- The argv-leak regression test still passes
- Pinned akm-cli version supports the new shape (CI installs that version)

## Dependencies

- Blocked on: a stable akm-cli release that includes commits `eff3e37` and `c50f9f4`

## Out of scope

- Phase 2 of #388 (drop the `vault/user → /etc/vault` compose mount, entrypoint sourcing) — separate follow-up

## References

- akm upstream issue (resolved): itlackey/akm#361
- akm commits: itlackey/akm@eff3e37, itlackey/akm@c50f9f4
- Implementing PR (Phase 1 with workaround): #404

itlackey:
## Summary

Closes two related deferred items:
- **Issue #406**: switch akm secret-store writes to the stdin pattern (akm 0.8.0-rc2 made stdin the default and removed the positional value form)
- **Phase 2 of issue #388**: drop the OP_HOME/vault/user → /etc/vault compose mount; entrypoint sources akm secret-store path; delete vault/user/

## Why

PR #404 implemented Phase 1 of #388 by mirroring vault/user/*.env into an akm secret store via direct .env file writes (bypassed the CLI to avoid the argv-leak vulnerability). Compose still mounted vault/user → /etc/vault for runtime env. With akm 0.8.0-rc2 (PR #408) shipping the stdin pattern, both the bypass and the mount can be retired.

### Part 1: stdin migration (#406)
- `packages/lib/src/control-plane/akm-vault.ts`: new helper spawns akm with the appropriate subcommand and writes the value to stdin — value never crosses argv
- Same 2s `Promise.race` timeout from PR #404's hang fix is preserved
- The delete helper switched to the akm unset subcommand (already argv-safe)
- The direct .env file write code path is deleted
- Regression test verifies values flow via stdin (never argv)

### Part 2: Phase 2 mount removal (#388)
- `.openpalm/stack/core.compose.yml`: vault/user → /etc/vault mount removed from assistant
- `core/assistant/entrypoint.sh`: sources the akm secret-store path before launching OpenCode
- Install/upgrade flow:
  - On `openpalm upgrade` from 0.10.x: after the existing mirror runs, delete vault/user/
  - On fresh `openpalm install`: don't create vault/user/ at all
- `load_vault` assistant tool: fallback to /etc/vault/user.env removed (the mount no longer exists)
- Updated regression test for the 0.10.x → 0.11.0 upgrade migration

### Part 3: NON-CHANGES
- vault/stack/stack.env and vault/stack/guardian.env remain operator-managed
- Guardian's HMAC secrets stay in vault/stack/guardian.env (env_file with hot reload)

## Acceptance

- [x] `bun run test` — 877/0 pass
- [x] `bun run admin:check` — 0 errors / 0 warnings
- [x] argv-leak regression test asserts values flow via stdin only
- [x] 0.10.x → 0.11.0 migration test asserts vault/user/ is gone post-upgrade
- [x] Tests updated to reflect Phase 2 contract (user.env no longer expected)

## Breaking change for downstream tooling

Anyone with custom shell scripts reading OP_HOME/vault/user/user.env directly should switch to `akm vault path vault:user`.

## Refs

- Closes #406
- Completes Phase 2 of #388
- Refs #409 (deferred-items follow-up)

### DIFF
--- a/.openpalm/stack/core.compose.yml
+++ b/.openpalm/stack/core.compose.yml
@@ -3,15 +3,22 @@
 # This file defines the core infrastructure. Addon services are added
 # via compose overlays in stack/addons/.
 #
-# Docker Compose reads variable values via --env-file for three env files:
+# Docker Compose reads variable values via --env-file for two env files:
 #   vault/stack/stack.env    — system config, API keys, OP_CAP_* capabilities
-#   vault/user/user.env      — optional user-managed custom vars
 #   vault/stack/guardian.env  — channel HMAC secrets (loaded by guardian)
 #
-# Directory model (v0.10.0):
+# User-managed env secrets live in the akm `vault:user` store at
+# data/stash/vaults/user.env (akm-cli >= 0.8.0 layout). The assistant
+# entrypoint sources that file at startup via `akm vault path vault:user`
+# — see Phase 2 of #388 / closed by #406. The legacy
+# vault/user/user.env compose env_file was retired; on upgrade it is
+# migrated into akm and deleted.
+#
+# Directory model (v0.11.0):
 #   ~/.openpalm/config/       — user-editable, non-secret configuration
-#   ~/.openpalm/vault/stack/   — system-managed secrets (stack.env)
-#   ~/.openpalm/vault/user/    — user-managed runtime files (user.env)
+#   ~/.openpalm/vault/stack/   — system-managed secrets (stack.env, guardian.env)
+#   ~/.openpalm/vault/user/    — operational files only (apprise.yml, gws/gcloud/mgc auth dirs)
+#   ~/.openpalm/data/stash/   — akm-cli stash (skills, vaults, etc.) — vault:user lives here
 #   ~/.openpalm/data/          — service-managed persistent data
 #   ~/.openpalm/logs/          — consolidated audit/debug output
 
@@ -95,7 +102,10 @@ services:
       EMBEDDING_API_KEY: ${EMBEDDING_API_KEY:-}
       # Capability resolution (used by entrypoint.sh to drop unused provider keys).
       OP_CAP_LLM_PROVIDER: ${OP_CAP_LLM_PROVIDER:-}
-      # Google Cloud credentials (file lives in vault/user, mounted at /etc/vault)
+      # Google Cloud credentials (file lives in vault/user, mounted at /etc/vault).
+      # NOTE: the /etc/vault mount no longer carries `user.env` (Phase 2 of #388
+      # routed that through akm `vault:user`). The mount remains because gws,
+      # gcloud, and mgc rely on it for their auth directories below.
       GOOGLE_APPLICATION_CREDENTIALS: /etc/vault/gcloud-credentials.json
       CLOUDSDK_CONFIG: /etc/vault/.gcloud
       # Microsoft Graph CLI credential config
--- a/bun.lock
+++ b/bun.lock
@@ -134,6 +134,9 @@
         "dotenv": "^16.4.7",
         "yaml": "^2.8.0",
       },
+      "devDependencies": {
+        "@types/bun": "^1.0.0",
+      },
     },
     "packages/scheduler": {
       "name": "@openpalm/scheduler",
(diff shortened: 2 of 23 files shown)


itlackey:
## Phase 2 admin vitest fixes (commit d3adc442)

Aligned the 10 stale admin vitest assertions with the Phase 2 akm-only contract from f9e0451b.

### Production code
No changes — `buildEnvFiles` (lib) and the user-vault route already implement the Phase 2 contract correctly. Only test specs were stale.

### Test changes (4 files, 10 failing -> 0)

**`config-persistence.vitest.ts`** — 3 tests in `buildEnvFiles`
- Renamed and rewrote to assert the env-file list contains only `stack.env` + `guardian.env`. Added explicit assertions that even when a legacy `vault/user/user.env` lingers on disk it must NOT be promoted to a compose env_file (compose interpolation would shadow akm-sourced values).

**`ensure-secrets.vitest.ts`** — 2 tests in `ensureSecrets`
- "seeds vault env files": now asserts `vault/user/` directory exists (it still backs operational non-env files like `apprise.yml`, gcloud creds) but `user.env` is NOT seeded.
- "applies strict permissions": now checks the `vault/user/` directory mode (0700) instead of a non-existent `user.env` file.

**`secrets.vitest.ts`** — 1 test "creates vault directory if missing"
- Same correction: directory created, file not seeded.

**`routes/admin/secrets/user-vault/server.vitest.ts`** — 4 failing tests rewritten + 2 new
- The route now operates on the akm `vault:user` store, not on `vault/user/user.env`. Rewrote tests to use `vi.mock("@openpalm/lib", ...)` over the akm helpers (`ensureAkmUserVault`, `readAkmUserVaultFile`, `writeAkmVaultKey`, `deleteAkmVaultKey`) backed by an in-memory `Map`.
- Why mocks: the real helpers spawn the `akm` binary via `Bun.spawn`, which is unavailable inside vitest's Node worker pool (root cause of the original `Bun is not defined` error in the route). Real-binary E2E coverage already lives in `packages/lib/src/control-plane/akm-vault.test.ts` under `bun test`.
- Added explicit 503 coverage for GET/POST/DELETE when akm is unavailable so the `akm_unavailable` error path is exercised in CI.

### Verification
- `bun run admin:test:unit` -> 505/505 pass (was 492/503)
- `bun run test` (root, non-admin) -> 877/0 (unchanged)
- `bun run admin:check` -> 0 errors, 0 warnings

PR #531, opened 2026-06-24
Retire the v prefix everywhere + 0.12 finalization cleanup

itlackey:
Coordinated cutover for the **0.12.41 closeout**: version tags are now **bare semver (X.Y.Z)** across the whole platform — the `v` prefix is gone going forward. Runtime and release workflows flip together so pulls stay coherent. Also folds in the 0.12 finalization cleanup (dead code, doc accuracy).

## v-prefix retirement

**Write side → bare:**
- Docker images publish `openpalm/<svc>:X.Y.Z` (`release.yml` + `publish-voice.yml` meta tags + existence guards).
- Git summary tag is bare `X.Y.Z` (was `vX.Y.Z`) for `platform`/`all` units.
- Runtime writes bare `OP_*_VERSION` / imageTag — dropped `formatForDocker`; `env.ts` `resolveRequestedImageTag` and `lifecycle` imageTag return `normalizeVersion` output.
- CLI self-update + UI ui-version route request bare refs.

**Read side stays v-tolerant (clean break, no image backfill — rely on `:latest`):** `normalizeVersion`, the Docker Hub resolver (`versions/latest`), the self-update redirect regex, and `groupReleasesByUnit`'s legacy pattern all still accept a leading `v`, so pre-cutover images/releases keep working.

## Consolidation + dead code
- Deleted the unused client-side `version-compare.ts` duplicate; the one server route that compares semver now imports `compareComparableVersions` from lib.
- Pruned dead lib exports: `formatForDocker`, `ensureCoreCompose`, `seedAssistantPersonaFiles` (zero production callers after the 0.12.34–0.12.40 reconcile consolidation).

## Docs
- Fixed stale `.openpalm/` → `packages/skeleton/` references (setup-guide, installation, system-requirements, environment-and-mounts).
- Removed the deleted `/admin/upgrade` endpoint from api-spec.
- Updated release-architecture tag scheme; CHANGELOG Changed/Removed entries.

## Verification
Gate green: **lib 646/0 · cli 52/0 · guardian 179/0 · ui:check 0/0 · ui:test:unit 860/0.**

⚠️ **Workflow YAML changes are verified by reading + YAML-parse + the test suite, not by an actual release run** — they prove out fully only when 0.12.41 is cut. The transformation is a faithful `v`→bare rename preserving existing structure (including the pre-existing quirk that the `platform` unit's bare summary tag carries no release; self-update resolves cleanly only off `all` releases — left as-is, not silently changed).


itlackey:
Thanks @copilot — all four comments were the same valid issue: stripping `v` from an **explicit** version pin broke pinning a pre-0.12.41 legacy image/release that only exists under `vX.Y.Z`.

Fixed in c984340d by making the explicit-pin paths **pass-through (verbatim)** rather than normalizing:

- `resolveRequestedImageTag()` (env.ts) — validates + returns the ref as-typed. `0.9.0` → `0.9.0`, `v0.12.40` → `v0.12.40`. `main`/invalid → null.
- CLI `self-update --version` — trims only and passes the tag straight to the GitHub download URL, so a legacy `vX.Y.Z` release stays fetchable.
- Tests updated to assert verbatim behavior (`main.test.ts`): legacy `v` preserved, bare stays bare.

This addresses the concern without reintroducing `v` anywhere the platform controls: the platform's own default tag flows through `PLATFORM_VERSION`, which is always bare. Only an explicit user-supplied legacy pin keeps its `v`. CHANGELOG updated to spell this out.

Gate after the fix: lib 646/0, cli 52/0, ui:check 0/0.

PR #582, opened 2026-07-25
Make the assistant reachable from the network, and replace presets with capability toggles

itlackey:
Opening the assistant from a phone has never worked. It fails three times in sequence; this fixes all three, hardens the credential boundary that becomes load-bearing once the UI is published, and replaces the network-access preset enum with composable toggles.

Supersedes #583 (design doc), merged in here.

## Why it never worked

| # | Failure | Fixed by |
|---|---|---|
| 1 | `Host` → **400 `invalid_host`** — allows only loopback, and runs as the *first* statement in `handle`. Its escape hatch never reached the container. | commit 3 |
| 2 | **CORS preflight fails** — the browser calls OpenCode cross-origin; OpenCode grants only loopback origins. | commit 4 |
| 3 | **401 from OpenCode** — the seeded connection hardcoded `auth: { mode: "none" }` regardless of `OPENCODE_AUTH`. | commit 4 |

An earlier revision of this branch fixed **only #2**. Withdrawn — it would have replaced a CORS error with a 400 and called it fixed.

## Commits

**1 — Port inversion and stale docs.** Compose fallbacks carried the *retired* pair (assistant `3800` / UI `3810`). The comment justified it as a pre-migration guard, but `migrateLegacyDefaultPorts()` runs before Compose on all four deploy paths, so it was unreachable and only inverted the layout for the manual runbook. Also: the voice container was documented as serving a browser UI — **it has no UI**, it serves an OpenAI-compatible API through the `/voice` proxy.

**2 — Login throttle + session signing key.** Both pre-existing; both gate publishing the UI. `/api/auth/login` had no attempt counter. And session tokens were `<expiresAt>.<HMAC(expiresAt, password)>` — signed with the master credential over a message shipping in cleartext in the same cookie, crackable at one HMAC-SHA256 per guess. Now `HMAC(serverKey, sha256(password))`.

⚠️ **Existing sessions are invalidated once on upgrade.** Users sign in again.

**3 — Host allowlist scoped.** The check asserts *"this service is loopback-only."* Permanently true for the admin-capable host process; false for a published container UI. Relaxation requires **both** a container marker and a non-loopback bind — DNS rebinding targets loopback-bound services, so a stray env var must never weaken the host process.

**4 — Same-origin `/oc` proxy.** Not a new architecture — **finishing one**. `/voice` already worked this way; the speak/transcribe relays were retired. OpenCode was the last browser-direct service, and the CORS module, loopback rewrite, mixed-content policy and second password were all consequences of that single gap. The entrypoint seeds a **root-relative `/oc`**, which the browser resolves against the origin it actually loaded — precisely what a process writing `runtime-config.json` at startup cannot know.

**5–6 — Toggles replace presets.** A four-value enum could not express *"network access on, guardian on for Slack, OpenAI API on, no screening"*, and every preset it *could* express had to be reverse-engineered back out of the env, because intent was only ever stored as its own consequences. Reading a toggle replaces inferring one, so nothing reads as "custom" and there is no drift to diagnose.

The wizard now asks **one question** — can other devices use the assistant. Guardian toggles appear only once a guardian-backed integration is selected; direct assistant exposure sits under Advanced and names its plain-HTTP risk.

## Deleted

- `network-preset.ts` + suite (~320 lines), `detectNetworkPreset`, `validateNetworkPresetEnv`, `collectNetworkExposureWarnings`, `collectBindAddressWarnings`' per-variable matrix
- `ui-cors-origins.ts`, the `OP_UI_CORS_ALLOWED_ORIGINS` grant, the entrypoint's whole `--cors` assembly, `OP_UI_DEFAULT_ASSISTANT_URL`, `OP_UI_HOST_PORT` / `OP_HOST_UI_PORT`
- **The cascade.** `${OP_X_BIND:-${OP_BIND_ADDRESS:-127.0.0.1}}` meant "unset" inherited for four listeners and meant loopback for one, *in the same file*. `OP_BIND_ADDRESS` is retired.
- The admin tab's raw "LAN exposure" checkbox — it published OpenCode while leaving the UI on loopback, so the switch labelled "LAN exposure" did not expose the front door.

Two exposure surfaces collapse: **voice** binds to a `127.0.0.1` literal (a global bind could previously publish a GPU inference API as a side effect), and the guardian's OpenAI-compatible listener gets **one** host port instead of two mapping to the same container port.

`GUARDIAN_DIRECT_INGRESS` is derived rather than hand-set, so a published guardian port can no longer point at a listener that 404s — the pairing route previously told operators to edit `stack.env` and restart the guardian when a paired device failed.

## Testing

**1589 UI tests pass**, `svelte-check` clean across both packages, Biome clean. Lib at its pre-existing 12-failure baseline (verified against a stashed tree — all uid/gid tests assuming a non-root user). CLI/electron failure sets diff clean against baseline.

The payload round-trip walks **all 16 toggle combinations**; the preset version could round-trip only the four it named.

## Not verified from a real device

This is the honest gap, and it is what let three sequential blockers hide behind each other. Every layer is unit-tested, but nobody has loaded the container UI from a phone. **That should happen before merge** — enable network access, open it from another device, confirm chat streams.


fwdslsh-dev:
## Review from #579 (docs branch) — no conflicts, three non-blocking findings

Reviewed while checking interaction with #579. **The two PRs share zero files** and branch from the same `main` (`b28b41a`), so no textual conflict is possible in either order.

**Suggested order: this PR first.** #579 cites line numbers inside `core.compose.yml`, which this PR shifts by +12 above the volumes block (`:149-151` → `:161-163`). Merging #582 first lets those references be refreshed before #579 lands. Both PRs also touch `docs/technical/environment-and-mounts.md`, but at lines 82 vs 99+ — far enough apart to merge cleanly regardless.

### The core work checks out

The compose-fallback inversion is a real latent footgun, and the argument for why the old guard was unreachable (`migrateLegacyDefaultPorts()` running before compose on all four deploy paths) holds up against the source.

I specifically verified the load-bearing safety claim, since this PR newly binds the admin UI to `0.0.0.0` under `home-open`: the login wall **is** fail-closed. `containers/assistant/entrypoint.sh:288-293` — with no usable password, every HTML navigation dead-ends at `/login` and no session can be minted. So this is not an unauthenticated LAN exposure of the admin surface. Worth calling out explicitly in release notes anyway, since "open access" now covers a surface it previously didn't.

### Findings

**1. Existing installs get a false "Home network" label — reproducing the exact dead-end this PR fixes.**

`detectNetworkPreset` drops the `!uiExposed` condition (`network-preset.ts:231`), so a legacy install with the UI still pinned to loopback reports `home-password`/`home-open`. The wizard's new copy then points that user at `http://openpalm.local:3800`, which is not bound on their box.

The PR body says "any re-save upgrades them in place" — but nothing prompts a re-save. There is no bind-address migration (only `migrateLegacyDefaultPorts` for ports), and `detectNetworkPreset` returns a bare preset string with no drift signal at either consumer (`api/setup/current-config/+server.ts:154`, `api/host/stack/+server.ts:65`).

The permissive detection is the right call — relabelling those installs "custom" would be worse. The gap is that nothing closes the loop afterward. Either a migration that rewrites the row, or a `needsResave` flag the UI can surface, would finish the fix for users who already have the broken state.

**2. `mergeUiCorsOrigins` never prunes, so the var grows without bound.**

Additive-by-design is right — silently deleting an operator's reverse-proxy origin would be the worse failure. But every DHCP re-IP permanently adds an origin, and nothing ever removes one. Pruning only entries matching the *derived shape* (`http://<ipv4>:<uiPort>`) while preserving everything else keeps the operator-safety property and bounds growth.

**3. The new `mdns.ui` entry reports desired state, not observed.**

`resolveMdnsStatus` is fed `readStackEnv()` — the file just written (`api/host/stack/+server.ts:53,61`) — so `ui.advertised` reflects the config, not the running binding. `AssistantTab.svelte` now renders that as the primary clickable front-door link, so between saving a Home preset and recreating the container the admin UI shows a live link that will not connect.

This isn't introduced here — the `assistant` entry already had it — but it lands on a more prominent surface. Flagging rather than requesting a change; it's the same root cause as the mDNS/observed-state items in the remediation plan on #579.

### Also

`ui-cors-origins.ts` is well covered — the DHCP-invalidation, idempotence and operator-preservation cases are all pinned, which is more than the module strictly needed.

Reviewing this also caught two errors in #579's own docs (a wrong `entrypoint.sh` citation and a missed inverted mount row); both are fixed on that branch.

---

itlackey:
@codex do a thorough review of this PR, insure it is simplifying and improving the overall quality and robustness of the code and improving stability and UX for the users

Issue #581, opened 2026-07-24
Reduce Container Storage Growth


itlackey:
## Summary

Container runtime storage grows unnecessarily over time because dependency installations, package-manager caches, generated artifacts, and rollback copies are retained in persistent container storage.

The main contributors are:

- Mutable dependency updates during container startup.
- Multiple versions and platform variants retained in package caches.
- Duplicate `node_modules` trees and caches across services.
- Rollback or archive workflows copying regenerable runtime artifacts.
- Insufficient visibility into which container paths consume storage.

### 1. Make runtime installs deterministic

- Replace assistant startup-time `bun update` with idempotent `bun install`.
- Pin assistant tool dependencies to reviewed exact versions.
- Follow the existing guardian and portal exact-pin installation pattern.
- Move dependency upgrades into the image build or release workflow.
- Add CI checks preventing mutable startup-time dependency updates.

### 2. Separate durable data from regenerable storage

- Keep package-manager caches, browser caches, generated artifacts, and tool `node_modules` outside durable service data where practical.
- Use explicitly named, independently purgeable cache volumes or the documented ephemeral cache location.
- Document which container paths are safe to remove and regenerate.
- Ensure cache cleanup never removes secrets, knowledge, active sessions, or other durable state.

### 3. Prevent backup duplication

- Preserve the existing targeted lifecycle backup and rollback allowlists.
- Update `docs/backup-restore.md` to describe curated durable backups instead of recursive copies of all runtime data.
- Exclude `.cache`, `.bun`, `.npm`, `tools/node_modules`, browser automation caches, and other regenerable artifacts from repository-owned backups.
- Add regression tests proving generated dependency artifacts are excluded from backup and rollback paths.
- Add retention and size visibility for old rollback and backup snapshots.

### 4. Add storage diagnostics

- Provide a host-side storage report showing the largest OpenPalm container data paths.
- Separate durable state, dependency trees, session databases, logs, and caches in the report.
- Show the effective Compose project, compose files, profiles, and persistent paths used by the running stack.
- Warn before upgrades, restarts, or backups when container storage exceeds a configurable safety threshold.
- Recommend supported cleanup actions instead of requiring manual filesystem inspection.

## Relevant Files

- `containers/assistant/entrypoint.sh`
- `containers/assistant/tools/package.json`
- `containers/guardian/entrypoint.sh`
- `containers/portal/start.sh`
- `packages/lib/src/control-plane/backup.ts`
[...]

Issue #577, opened 2026-07-22
Code-review follow-ups from #574: activation veto contract + redundant IndexedDB read


fwdslsh-dev:
Two non-urgent code-quality items surfaced in the review of #574 (`refactor/setup`). Neither is a live bug; grouping them so they can be picked up later.

## 1. `emitConnectionActivated` vetoes activation via a fragile `=== false` return

`packages/ui/src/lib/connection-events.ts` (~line 76): the activation-refusal path is an implicit "listener returned the literal `false`" check, and `ActivationListener`'s return type was widened to `unknown`:

```ts
type ActivationListener = (connectionId: string) => unknown | Promise<unknown>;
// ...
if ((await listener(connectionId)) === false) {
  throw new Error('Connection activation was refused by a listener.');
}
```

Today this is safe: only `chat.onEndpointChanged` is meant to veto (returns `boolean`), and the other registered listener (the chat page's session-navigation handler) returns `undefined`. But the contract is a footgun — any future `onConnectionActivated` subscriber written as a predicate-style function that happens to return `false` would silently abort (and roll back) the activation, with no type-level signal that a falsy return has that side effect.

**Suggested cleanup:** make the veto explicit — e.g. a dedicated sentinel/typed result (`{ refuse: true }` or a named symbol) instead of overloading `false`, and narrow the listener return type accordingly so a plain boolean-returning listener can't accidentally veto.

## 2. `pickStorage` does a redundant `getAll()` on first storage access

`packages/ui/src/lib/connections/boot.ts` (~line 60): the IndexedDB availability probe in `select()` runs `persistent.getAll()` purely to decide persistent-vs-memory and discards the result; the wrapper's first `getAll()` then issues a second `getAll()`:

```ts
selected ??= persistent.getAll().then(
  () => { storageMode = 'persistent'; return persistent; },
  () => { storageMode = 'session-only'; return memory; }
);
// ...
getAll: async () => (await select()).getAll(),
```

So the first `getConnectionStorageMode()` / first store access performs two reads. Low impact (connection sets are small), but the probe throws away entries it just read. **Suggested cleanup:** have `select()` surface the probe's result so the first `getAll()` can reuse it, or probe with a lighter open-only check.

fwdslsh-dev:
Captured in the public-seams review (**#584**) as **Section 9 (U1–U2)** of `docs/public-seams-review.md`. Both verified against current code:

[...]

PR #584, opened 2026-07-26
Public-seams review + code-verified implementation of the findings

fwdslsh-dev:
## What this PR now contains

This started as a **documentation-blind public-seams review** (`docs/public-seams-review.md`) and now also carries a **code-verified implementation** of the findings, done in three waves with a test-first → fix → verify loop and per-batch adversarial review. It also folds in issues **#581** (container storage growth) and **#577** (UI code-quality) — see Sections 8–9 of the doc.

The review doc is the authoritative spec; every code change maps to a finding ID (G1, E2, S3, …).

## Implemented — unit-tested & green in CI-equivalent runs
- **Lifecycle/CLI:** C1 (`uninstall --purge` removes `state/`+`system/`), C3 (password live-read, no-restart + `reset-password`), D1 (friendly day-2 docker errors), B5 (crontab win32 guard)
- **Security/portals:** G1 (delegated secrets relocated out of the assistant stash → `private/secrets`, guardian `auth.json` via compose secret, non-destructive idempotent migration, env:user no longer sourced into the OpenCode server), G3 (portal default-deny + explicit `"*"` opt-in), G6 (jailbreak-weight retune), G7 (empty-key → 401)
- **Supply chain/versioning:** E1 (image-tag pin to `PLATFORM_VERSION`, manifest-guarded), F1 (rootless chown on install), F2 (`dockerBin()` Podman seam), S5 (backup lifecycle: mtime prune, atomic writes, wired space-guard, external destination)
- **Diagnostics/storage:** C2+S6+S7+S8 (unified `openpalm doctor` — env checks, disk-headroom preflight, image/volume retention, storage report + `--clean-caches`), S3 (OpenCode session/DB retention + `VACUUM`, bun:sqlite-tested), S4 (portal session reuse + guardian orphan reconciliation)
- **Distribution:** A1 (pure-node npm bootstrapper; dependency-free published package), U1/U2 (activation-veto contract, redundant IndexedDB probe)

**Test status:** lib 1014✓/12 (pre-existing root-uid env), cli 170✓/3 (pre-existing env), guardian+portals 533✓/0, ui 1531✓/0, scripts 55✓/0. No new regressions.

## Implemented — structural-only, **must be validated in a real environment before merge**
These could not be executed in the build sandbox (no image builds, no live stack, no CI runner, no Windows/PowerShell). They are parse/lint/structural-verified only:
- **E2/S2/S1** — image-baked tools (remove runtime installs, bake skeleton, exact-pin tools + plugin, drop tools bind mounts, relocate caches out of durable `data/`). ⚠ Needs: image build; live + **rootless** stack boot; the new `assistant-cache`/`guardian-home-cache` volumes added to `volume-ownership.ts` `SERVICE_NAMED_VOLUMES`; an **upgrade/migration step** for existing named volumes (Docker only seeds a named volume on creation, so pre-existing installs keep stale content until recreated).
- **G1 compose/entrypoint parts + the migration** — validate the secret migration on a real `OP_HOME` upgrade (fresh / partial / already-migrated) and confirm the agent can no longer read `env:user` via `env`/`printenv` on a live assistant.
- **E4** — cosign keyless signing + in-CI verify: validate on a real prerelease CI run.
- **E3** — opt-in voice release tie-in.
- **B1–B4** — PowerShell installer (checksum, PATH persistence, drop PS7 gate, arm64). Validate on real Windows (5.1 + 7). B4's `windows-arm64` build has `continue-on-error` because Bun's cross-compile support for that target is unverified.

## Not changed (by decision)
- **G4** (self-approval of permission prompts) — **by design**, withdrawn.
- **B6** (Windows secret ACLs) — **accepted limitation**, deferred.
- **G2** (opencode denylist patterns) — **already fixed** on the base; regression test extended.

## Notes for the reviewer
- Commits are correctly attributed to `Claude <noreply@anthropic.com>` but show **Unverified** — the sandbox has no commit-signing key available; nothing else can be done here.
- Doc-sync follow-ups remain (a few stale comments referencing removed boot-time-install behavior) — non-blocking, flagged in commit messages/residual-risk notes.


itlackey:
@codex do a thorough review of this PR

fwdslsh-dev:
Thanks @chatgpt-codex-connector — worked through all 7 findings. The four contained ones are fixed in this PR (`017ef5b`), each with tests; the two architectural ones are tracked as follow-up issues since they need a redesign + real-stack validation rather than a contained patch.

**Fixed here (`017ef5b`)**

| # | Finding | Fix |
|---|---------|-----|
| P1 | Purge the new private secrets tree | `uninstall --purge` now removes `private/` (added `resolvePrivateDir()` to lib + the purge dir list). Kept the credentials in the deliberately-out-of-`knowledge/` `private/` tree per the §G1 delegated-secrets design (the agent must never reach them) and brought it into the purge/ownership lifecycle, rather than moving them back under the agent-reachable `knowledge/secrets`. Test asserts a seeded `private/secrets` credential is gone after `--purge`. |
| P1 | Verify every service image before pinning | Setup now probes `assistant`/`guardian`/`portal` **independently** and falls back to `latest` per missing image (voice unchanged). New test proves a lagging guardian/portal stays on `latest` when only `assistant:PLATFORM_VERSION` is published. |
| P1 | Stamp baked adapters during portal releases | Added `stampPortalBakedManifest()` (exact pins for a reproducible build) wired into `UNITS.portals.stamp`, so a `unit=portals` release stamps `containers/portal/tools/package.json` to the just-published version, not just the (now-inert) seed. |
| P2 | Wire DB maintenance into a production path | Wired `runOpenCodeDbMaintenance` into a confirm-gated `openpalm doctor --reclaim-db`. Extended it to accept a `null` client for **file-only** reclamation (checkpoint + conditional VACUUM) — the disk-reclaim stage needs the assistant *stopped*, while a live REST session-delete needs it *running* (opposite preconditions), so `doctor` uses the file-only path and a locked DB surfaces as an actionable error. |

**Tracked as follow-ups** (architectural — need redesign + live-upgrade/stack validation)

- #585 — **Refresh named volumes when baked artifacts change** (P1). `assistant-artifacts`/`guardian-cache`/`portal-cache` obscure the new image's `/opt/openpalm` on upgrade; needs a volume recreate/migrate strategy verified on a real upgrade.
- #586 — **Reconciliation deletes active sessions / drops pending rows** (P1 + P2, `reconciliation.ts:70` + `state-db.ts:337`). Both stem from age/count retention ignoring lifecycle state; needs a redesign around verified session state.

CI is green on the contained changes (biome lint clean; new paths covered by unit tests). The pre-existing root-uid ownership/rollback test failures in the sandbox are environmental and unrelated to this diff.

---

PR #587, opened 2026-07-27
0.13.0 release blockers: #586, #585, #581, #577

### LINKED ISSUE #577: Code-review follow-ups from #574: activation veto contract + redundant IndexedDB read

### LINKED ISSUE #581: Reduce Container Storage Growth

### LINKED ISSUE #585: E2 upgrade path: image-baked named volumes are not reseeded on upgrade
## Summary

Follow-up to the Codex review on #584 (finding: _"Refresh named volumes when baked artifacts change"_, P1). Tracked separately because the fix is architectural — it needs a volume recreate/migrate strategy plus real-stack upgrade validation, not a contained code change.

## The problem

The image-baked-only model (§E2/§S2) serves `/opt/openpalm` (baked UI, skeleton, and tool trees) out of named volumes:

- `assistant-artifacts` → `packages/skeleton/system/stack/core.compose.yml`
- `guardian-cache`, `portal-cache` → `packages/skeleton/system/stack/portals.compose.yml`

Docker only propagates an image directory's contents into a **freshly created, empty** named volume. On any upgrade from an existing install these volumes are already populated, so mounting them **obscures** the new image's `/opt/openpalm` tree and Docker does **not** reseed them ([Docker volume docs](https://docs.docker.com/engine/storage/volumes/#mounting-a-volume-over-existing-data)).

Because #584 also removed the nested tools bind mounts and the assistant/portal runtime update paths, an upgraded container keeps running the artifacts captured in the **old** volume — the newly baked UI, skeleton, and tools never take effect. The user pulls a new image, restarts, and silently runs stale code.

## Why it's not fixed in #584

The two candidate fixes both carry risk that needs live-upgrade validation:

1. **Recreate on upgrade** — `docker compose down -v` / targeted `docker volume rm` for these specific volumes before `up`, so Docker reseeds them from the new image. Must be scoped to exactly the baked-artifact volumes and must not touch data volumes (`data/*` binds are host-side, but any future named data volume must be excluded).
2. **Version-stamped volume names** — suffix the volume name with the image version so a new image always mounts a fresh, empty volume (old ones become orphans that `doctor --clean-docker` reaps). Avoids a destructive down step but changes the volume-naming contract.

Either way this needs a real image-build + upgrade pass to confirm the reseed actually happens and nothing that should persist is lost.

## Acceptance criteria

- Upgrading an existing install to a new image serves the **new** baked `/opt/openpalm` contents (UI, skeleton, portal/guardian tools), verified on a live upgrade.
- No host-side persistent data (`data/*`) is destroyed by the migration.
- Orphaned old volumes (if version-stamped) are reaped by `openpalm doctor --clean-docker`.

## References

- PR #584 review thread: https://github.com/itlackey/openpalm/pull/584#discussion_r3654842637
- `docs/public-seams-review.md` §E2 / §S2

### LINKED ISSUE #586: S4: guardian reconciliation can delete active sessions and drop pending rows
## Summary

Follow-up to the Codex review on #584 (two related findings). Tracked separately because a correct fix requires redesigning the guardian's ownership/eviction retention around verified session lifecycle state, plus real-stack validation — not a contained code change.

## Finding A — active sessions deleted purely for being "old" (P1)

`packages/guardian/src/reconciliation.ts:70`

Once `GUARDIAN_OWNERSHIP_MAX_ROWS` is exceeded, every evicted ownership row is passed straight to the upstream `DELETE` callback with no check for whether the session is archived, stale, or currently active. Eviction is ordered by the original `created_at`, which is **never refreshed on reuse or access**, so a long-lived **active** portal conversation can be the oldest row and get deleted by the next reconciliation sweep simply because enough newer sessions were created.

**Impact:** an active Discord/Slack/API conversation's upstream session is destroyed mid-use.

## Finding B — pending (unreconciled) rows discarded under backlog (P2)

`packages/guardian/src/state-db.ts:337`

If the reconciliation queue grows beyond `GUARDIAN_EVICTION_LOG_MAX_ROWS` while the assistant is unavailable, or creation outpaces the 100-per-5-minute sweep, the cap query eventually deletes **unreconciled** rows once all reconciled rows are exhausted. Those session IDs have already lost their ownership records, so dropping their only queue entries recreates the permanent, undeletable upstream orphans this table exists to prevent.

**Impact:** the exact orphan-leak this table was built to stop, reintroduced under load.

## Direction

Both stem from age/count-based retention that ignores lifecycle state. A redesign should:

- Base ownership eviction on **verified session lifecycle** (archived/completed) rather than raw `created_at` age; never evict a row for a session known to be active/recent.
- Refresh the ordering key on access/reuse, or track `last_used` separately, so a live session is never the "oldest".
- Never discard **pending** reconciliation rows to satisfy the cap — apply backpressure, accelerate/enlarge the sweep, or prune only **completed** history.

## Acceptance criteria

- A long-lived active portal session survives an eviction sweep that crosses `GUARDIAN_OWNERSHIP_MAX_ROWS`.
- Under a reconciliation backlog exceeding `GUARDIAN_EVICTION_LOG_MAX_ROWS`, no pending row is dropped and no permanent upstream orphan is created.
- Regression tests cover both the active-session and backlog scenarios.

## References

- PR #584 review threads:
  - Active-session deletion: https://github.com/itlackey/openpalm/pull/584#discussion_r3654842657
  - Pending-row drop: https://github.com/itlackey/openpalm/pull/584#discussion_r3654842660
- `docs/public-seams-review.md` §S4

itlackey:
Closes the four open 0.13.0 milestone issues. Plan and recorded decisions: `.github/roadmap/0.13.0/final-four-plan.md`.

## Measured results

| Image | Before | After |
|---|---|---|
| assistant | 4.77 GB | **1.45 GB** (−70%) |
| guardian | 2.03 GB | **1.42 GB** (−30%) |

A default install goes from ~7 GB at rest (image + a duplicate ~2.2 GB volume copy) to 1.45 GB, so the documented minimum drops from **10 GB to 4 GB**. Upgrade pulls drop from ~3.3 GB to ~25 MB, and old/new images now share nearly every layer instead of coexisting in full.

## #585 — the fix is deletion, not machinery

The planned reseed module was withdrawn. An audit established that nothing under `/opt/openpalm` needs to survive container recreation: the assistant writes one file there per boot and regenerates it every time, the guardian's durable state (`state.db`, audit log, `auth.json`, bun cache) lives on host binds *nested under* the volume, and the portals only read. `guardian-cache` was introduced as an npm artifact cache for the boot-time installs that #584 deleted.

The stale volume was also *causing* the data pull: it failed the guardian's already-at-version check, so every upgrade re-fetched packages from npm. Deleting the three mounts dissolves the issue, reclaims ~2.2 GB per install, and lets the volume-ownership subsystem shrink to one entry. Retired volumes are reaped automatically on upgrade behind a closed, current-project-scoped list that can never match `assistant-persistent`.

## #581 — the remainder

- **S1 cache relocation** to pre-created `${OP_HOME}/cache/` host binds. The first attempt (`921412b1`) was reverted because nested named volumes made Docker create root-owned mountpoints; binds pre-created by `ensureHomeDirs` avoid that. Companions: excluded from backups and the space estimate, covered by `--purge`, purgeable via `--clean-caches`, and the filesystem contract amended.
- **S3 live wiring** — `doctor --sessions` and `--prune-sessions`, thin adapters over the existing lib client. A retention window is mandatory; root sessions are never deleted.
- **S5/S6/S8** — the prune preview now matches the prune (it could previously list survivors and omit casualties on a destructive prompt), bundle snapshots bounded, `OP_BACKUP_DIR` honored by doctor/install/purge, and the disk preflight extended to the lib apply path that UI-driven installs use.
- Image slimming per the recorded decisions, with removed tooling now installed on demand via a new skill.

## #586 — guardian schema v4

Lifecycle-aware eviction: `last_used_at` refreshed at the proxy choke point, a 24h active grace, upstream recency verification that restores rather than deletes an active session, and pruning restricted to reconciled rows so pending orphan records are never dropped.

## Verification

Built and measured both images. Confirmed in the running assistant that the offline embedding test still passes (`384 dims`) after pruning 479 MB of onnxruntime CUDA and wrong-platform binaries, that the removed tools are absent and the kept ones present, and that a version-only rebuild reuses the heavy layer. The rootless ownership smoke — the gate that failed the first S1 attempt — passes.

Gates green: 1902 tests, lint, svelte-check, `docker compose config`.

Two findings the build caught that review would not have: an initial layer split made the guardian image 790 MB *larger* (a recursive `chmod` copied the cached tools tree up into the next layer), and deleting the portal seed manifest orphaned a release-script function that stamped it.

Closes #586, closes #585, closes #581, closes #577.

https://claude.ai/code/session_016QgrghM4MY17eZi9GcsGPX


itlackey:
@codex review this PR and focus on docker best practices, subtle misalignment between the layers of the stack. Review the migration/upgrade code for any failures that would cause users to lose data in their existing installations.

itlackey:
Ahead of the Codex pass, here's a verified audit of the third concern — data loss in existing installations — since that's the highest-stakes part of this branch. Checked against the last shipped release and a real 80 GB install.

**The retired-volume reaper cannot destroy user data.** At `0.13.0-beta.13`, `core.compose.yml` describes `assistant-artifacts` as a *"Platform artifact cache — persists @openpalm/ui and @openpalm/skeleton across restarts so warm starts don't re-download"*, and the tools subpath is **overlaid by a host bind** (`${OP_HOME}/data/assistant/tools:/opt/openpalm/tools`), with the file's own comment saying so. So on a real pre-upgrade install the volume holds only re-downloadable npm artifacts. Anything a user or the agent installed went to one of two places the reaper never touches: the host bind (a file on disk), or `assistant-persistent:/opt/persistent`, the documented escape hatch — which is excluded by a closed list plus a negative test asserting no input can make the reaper target it.

Ordering is also safe: the reap runs **after** the new stack is confirmed up, on all three paths (`lifecycle.ts:333`, `deploy.ts:367` and `:389`), and failures are logged rather than thrown, so a reclaim failure can never strand an upgrade.

**The S1 cache relocation shadows, never deletes.** The new `${OP_HOME}/cache/{assistant,guardian}` binds hide the old in-`data/` cache paths in-container; the old bytes stay on disk untouched. On this machine that tree is 27 GB of `bun`, `ms-playwright`, `opencode`, `npm`, `gh` and `fontconfig` caches — a `find` for `*.db`, `*.sqlite*`, `auth*` or `*session*` under it returns nothing, so there is no durable state in the shadowed path. The legacy locations stay in the purge safelist, and a dry run against that install reports **28.2 GB reclaimable** via `openpalm doctor --clean-caches`.

Also checked: `cache/` is excluded from both the backup copy scope and the space estimate (so snapshots don't reacquire the bloat AC4 removed) and is covered by `--purge`; removing `data/*/tools` from `ensureHomeDirs` only stops *creating* those dirs — existing ones are left alone and documented as manually removable; and the `ui-*`/`skeleton-*` auto-prune is namespace-scoped with `-pre-rollback`/`-pre-update` recovery snapshots excluded outright.

One behavior worth recording rather than fixing: after the reap, **downgrading** to a pre-#585 OpenPalm recreates an empty `assistant-artifacts` and re-downloads `@openpalm/ui`/`@openpalm/skeleton` on first boot — slower, not lossy.

New coverage for this class landed in `scripts/upgrade-path-smoke.sh`: it upgrades homes seeded from *published* skeleton versions (npm keeps every release, so the fixture pool is free to maintain) and exercises the reaper against real Docker volumes, asserting `assistant-persistent` survives. Mutation-verified — adding `assistant-persistent` to the retired list, dropping `cache/` from `ensureHomeDirs`, and skipping the `rmSync` that removes a migrated secret each make it fail.

itlackey:
Thanks — two of these were real and are fixed in `4803c791`. Three I'm pushing back on with evidence.

**Fixed — P1 archived-only deletion.** Correct and the most valuable finding here. The predicate filtered on `parentID` and idle time only, so enabling the task would have deleted live branches that merely went quiet — strictly more destructive than the host-side path, whose `computeRetentionPlan` defaults `requireArchived` to true. Now requires `time.archived` (the field `toSessionRecord` reads). Verified against a fixture covering all four cases: unarchived-idle child excluded, recently-archived child excluded, root excluded, only archived-and-old selected.

**Fixed — P2 authentication.** Also correct, and it would have made the task silently do nothing rather than fail loudly: under `OPENCODE_AUTH=true` the first `curl -f` takes a 401 and exits non-zero. Now uses the same auth branch as the assistant healthcheck, reading the mounted secret, and errors explicitly if auth is on but the secret is unreadable.

**Not changing — P1 volume ownership.** The premise doesn't hold: Docker refuses to remove a volume referenced by *any* container, stopped included. Verified directly —

```
$ docker create --name vtest -v testproj_guardian-cache:/d alpine true
$ docker volume rm testproj_guardian-cache
Error response from daemon: remove testproj_guardian-cache: volume is in use - [fd92ed00...]
```

So "another stopped stack's persistent volume" is already protected, and the reaper surfaces that as a logged error rather than deleting. The residual is a *fully dangling* volume whose name collides on both the project name and one of three exact suffixes. The proposed label check cannot discriminate that case — a same-project-name collision carries the same `com.docker.compose.project` value, so the label matches too. Adding a check that returns the same answer for both cases would buy confidence without safety.

**Not changing — P1 compose-only cache bind.** The dependency is real but pre-existing and not introduced here: *every* bind in `core.compose.yml` requires `ensureHomeDirs` to have run — `config/akm`, `config/assistant`, `data/assistant`, `data/akm/cache`, `system/assistant`, `knowledge`, `workspace`. The cache bind is consistent with that contract rather than a new exception, and a restored backup already lacks `data/` for the same reason (lifecycle snapshots exclude it), so the manual-compose flow has always required an initialized home.

I did evaluate the obvious hardening — adding the cache dirs to `repairRootOwnedBindMounts` — and rejected it as a net regression: in `strict`/`deep` mode that helper repairs every candidate unconditionally with a recursive `chown` under a 30s timeout, and a populated cache is tens of GB (27 GB on the install I measured). It would trade a hypothetical for a likely timeout. Happy to revisit if the compose-only path becomes a supported first-class flow.

**Deferred — P2 Docker data-root headroom.** Legitimate gap, and out of scope for this PR. The S6 guard's contract is OP_HOME headroom; extending it to `DockerRootDir` changes what can *block* an install, which deserves its own change with its own tests rather than riding a release-blocking branch. Worth filing.

itlackey:
Filed the deferred P2 as #588 (disk-headroom preflight ignores Docker's data root when it is a separate filesystem).

Worth noting the scenario is demonstrably real rather than theoretical — on a maintainer machine the two live on different devices: `OP_HOME` on `/home` (837G avail) and Docker's root on `/` (648G avail). The dangerous inverse — roomy `OP_HOME`, nearly-full Docker root — is exactly what the guard would wave through today.

The issue records the decision that actually needs making: whether a critical Docker-root reading should block on the same `OP_DISK_HARD_BLOCK` terms as `OP_HOME`, plus the Docker Desktop case where `DockerRootDir` points inside the VM and a host `statfs` on it is meaningless.

itlackey:
CI is fully green on `4803c791` — all five checks pass and the PR is `CLEAN`.

Both newly-wired smokes ran and passed in CI, not just locally:

```
success  Guardian offline-boot smoke (--network none)
success  Upgrade-path smoke (historical homes + retired-volume reap)
```

That matters for the first one in particular: it guards the S.4 invariant (the trust boundary boots with zero registry access) and had been running nowhere, which is how its assertions silently rotted against a log line #584 reworded. It now actually gates.

Ready for review. Everything raised is resolved: two Codex findings fixed here, three answered with evidence above, and the deferred disk-headroom gap filed as #588.

Issue #461, opened 2026-06-08
Setup: detect + stop/recreate an existing OpenPalm stack instead of erroring


itlackey:
Install/setup errors with 'openpalm project is already running in docker' when a stack is up. Root: CLI `install.ts deployServices` does `compose up -d` with no pre-down / collision check / --force-recreate; UI wizard is only partially hardened; default project name 'openpalm' is shared across installs.
Fix: shared lib helper that detects our project (compose -p ps), force-recreates if it's ours, returns a structured 'foreign stack' error otherwise; call from BOTH CLI install and the wizard; always --force-recreate --remove-orphans on install/start; `down` before --force reinstall; write a deterministic OP_PROJECT_NAME. (lib/control-plane/docker.ts, lifecycle.ts; cli/commands/install.ts; ui/lib/server/setup-deploy.ts)

itlackey:
Shipped in 0.11.2 (released, npm `latest` + GitHub `v0.11.2`; merged to main).

PR #591, opened 2026-07-28
Refactor/deployment model

itlackey:
This pull request refactors how the UI and skeleton packages are handled in the build, packaging, and release process. The most significant changes remove the independent `@openpalm/skeleton` bake from the Assistant image, streamline the UI build to always use the local candidate, and introduce a dedicated workflow for publishing extension packages. Several references and scripts are updated to reflect new package locations and conventions, and unnecessary scripts and checks are removed.

**Build and Packaging Changes:**

* The Assistant Dockerfile now only bakes the compiled local UI candidate (`@openpalm/ui`) into the image, removing all logic and references for baking or installing `@openpalm/skeleton`. The UI is packed and installed directly from the local source during the image build. (`[[1]](diffhunk://#diff-8fa84988fff4d4af28e69931e74a3e2475e4db7cc05b9c3b3abf1ec72344adf5L206-R209)`, `[[2]](diffhunk://#diff-8fa84988fff4d4af28e69931e74a3e2475e4db7cc05b9c3b3abf1ec72344adf5L226-R231)`, `[[3]](diffhunk://#diff-8fa84988fff4d4af28e69931e74a3e2475e4db7cc05b9c3b3abf1ec72344adf5L243-R252)`, `[[4]](diffhunk://#diff-763596bc3b3aeb385c1d8dfda561736407874068c13d7deae45e4701091ee6faL1-L19)`, `[[5]](diffhunk://#diff-8bd634f3f155bac15fa1abd32de628046a155ebcc0b58a6ef073fd96cadaccc5L21-R28)`)
* The Docker build arguments and compose files are updated to clarify that the Assistant image always uses the local UI candidate and no longer fetches skeleton or UI from npm at build time. (`[[1]](diffhunk://#diff-bfb72f2c1c0b7deb52e32346fe2a5dadd2a489cedd430a2cc12e2ce57e7c2f51L33-R33)`, `[[2]](diffhunk://#diff-bfb72f2c1c0b7deb52e32346fe2a5dadd2a489cedd430a2cc12e2ce57e7c2f51L54-R46)`)

**Release and Extension Publishing:**

* A new workflow (`.github/workflows/publish-extensions.yml`) is added to independently validate, test, pack, and publish the Guardian and Portal extension packages. This workflow enforces version consistency and prevents accidental mismatches. (`[.github/workflows/publish-extensions.ymlR1-R78](diffhunk://#diff-4bbbe74dec2f0ef7c9b3597c6053132014c7cfd0f497c9dffaa3df0f9b66ee7aR1-R78)`)
* The main release documentation and package groupings are updated to reflect that only the main product publishes a public npm package, while UI, skeleton, and lib are now private workspaces. Extension packages (Guardian, Portal SDK, Discord, Slack) are published independently. (`[[1]](diffhunk://#diff-1955384557607f03dfcb60b699c7f19e34255bf88bb0988fc7edc4c76d60a827L8-R17)`, `[[2]](diffhunk://#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44cL164-R185)`)

**Project Structure and References:**

* All references to portal adapter packages are updated to their new locations under `packages/portal-*` instead of `portals/*`. This affects `.dockerignore`, documentation, and package groupings. (`[[1]](diffhunk://#diff-2f754321d62f08ba8392b9b168b83e24ea2852bb5d815d63e767f6c3d23c6ac5L18-R19)`, `[[2]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L43-R43)`, `[[3]](diffhunk://#diff-1955384557607f03dfcb60b699c7f19e34255bf88bb0988fc7edc4c76d60a827L8-R17)`)
* The CI workflow and contributing guide are updated to remove checks and references for the deprecated `ui-kit` package and to clarify the scope of type checks and tests. (`[[1]](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL131-L133)`, `[[2]](diffhunk://#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44cL102)`, `[[3]](diffhunk://#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44cL119-R118)`)

**Script and Consistency Checks:**

* The version consistency checks for setup scripts in CI are simplified, removing the requirement that setup script versions match the root version exactly. (`[.github/workflows/ci.ymlL303-R304](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL303-R304)`)

**Entrypoint and Runtime Behavior:**

* The Assistant entrypoint script and Dockerfile comments are updated to clarify that there are no runtime package installs or version checks for UI or skeleton—everything is baked into the image at build time. (`[[1]](diffhunk://#diff-3244d482b46f8611ab587c0e0c0f13ccebde17c602aaf310708b66d806b5db50L70-R72)`, `[[2]](diffhunk://#diff-3244d482b46f8611ab587c0e0c0f13ccebde17c602aaf310708b66d806b5db50L82-R81)`, `[[3]](diffhunk://#diff-8fa84988fff4d4af28e69931e74a3e2475e4db7cc05b9c3b3abf1ec72344adf5L226-R231)`, `[[4]](diffhunk://#diff-8fa84988fff4d4af28e69931e74a3e2475e4db7cc05b9c3b3abf1ec72344adf5L206-R209)`)

These changes collectively simplify the build and release process, reduce runtime complexity, and clarify the separation between product and extension packages.


.github/workflows/publish-extensions.yml
```diff
@@ -0,0 +1,78 @@
...
+          VERSION: ${{ inputs.version }}
+        run: |
+          node --input-type=module <<'NODE'
+          import { readFileSync } from 'node:fs';
+          const version = process.env.VERSION;
+          for (const file of ['packages/guardian/package.json', 'packages/portal-sdk/package.json', 'packages/portal-discord/package.json', 'packages/portal-slack/package.json']) {
+            const actual = JSON.parse(readFileSync(file, 'utf8')).version;
+            if (actual !== version) throw new Error(`${file} is ${actual}, expected ${version}`);
+           }
+           NODE
```
itlackey: **[Critical] This heredoc never terminates — the workflow fails on every dispatch.**

The closing `}` (line 48) and the `NODE` delimiter (line 49) are indented **11 spaces**, one deeper than the block scalar's base indentation of 10. After YAML strips the common indent, the delimiter line becomes ` NODE` (leading space), which does not terminate `<<'NODE'`. Bash reaches EOF ("here-document delimited by end-of-file") and feeds the leftover ` }` and ` NODE` lines into node as module source, which throws `ReferenceError: NODE is not defined` — so the "Validate extension versions" step exits 1 on **every** run, dry-run or live, even with perfectly matching versions. Reproduced locally.

Worth noting: the only executable check the test suite runs on workflows is `Bun.YAML.parse` (see `release-publish-dag.test.ts`), which cannot catch this — the YAML is valid; the embedded bash is not.

Fix: outdent both lines by one space to match the rest of the block.

---

containers/assistant/Dockerfile
```diff
@@ -240,40 +240,16 @@ ARG PLATFORM_VERSION
...
-    else \
-      echo "WARN: PLATFORM_VERSION unset at build time — no fallback @openpalm/skeleton baked (dev/local build only)" >&2; \
-    fi
+# The compiled candidate is copied from the checkout and packed locally. This
+# keeps live builds and coordinated dry runs on the same path without requiring
+# the candidate UI version to have been published first.
+COPY packages/ui /opt/openpalm/local-src/ui
+RUN mkdir -p /opt/openpalm/local-artifacts \
+    && (cd /opt/openpalm/local-src/ui && bun pm pack --destination /opt/openpalm/local-artifacts --quiet) \
+    && npm install --prefix /opt/openpalm/ui /opt/openpalm/local-artifacts/openpalm-ui-*.tgz \
```
itlackey, line 252: **[Critical] The assistant image build fails at this UI pack step.**

This stage `COPY`s only `packages/ui` — no workspace root, no `packages/lib` — but `packages/ui/package.json` declares `"@openpalm/lib": "workspace:*"` in `devDependencies`. `bun pm pack` then hard-fails with:

```
error: Failed to resolve workspace version for "@openpalm/lib" in devDependencies. Run bun install and try again.
```

Reproduced with bun 1.3.11 by copying the real `packages/ui/package.json` into an isolated directory laid out exactly as this Dockerfile does. Any `node_modules` symlinks from a host checkout would be dangling after COPY, so nothing in the build context rescues it. Every assistant image build — the release workflow's docker job and `bun run dev:build` via `compose.dev.yml` — dies at this `RUN`.

Fix options: also COPY the root `package.json` + `packages/lib/package.json` so the workspace resolves, or strip `devDependencies` from the copied manifest before `bun pm pack` (the pack only needs `build/` + runtime deps).

---

containers/portal/Dockerfile
```diff
@@ -1,20 +1,26 @@
...
+COPY containers/portal/workspace.package.json /opt/openpalm/local-src/package.json
+COPY packages/portal-sdk /opt/openpalm/local-src/packages/portal-sdk
+COPY packages/portal-discord /opt/openpalm/local-src/packages/portal-discord
+COPY packages/portal-slack /opt/openpalm/local-src/packages/portal-slack
+RUN mkdir -p /opt/openpalm/local-artifacts /opt/openpalm/tools \
+    && (cd /opt/openpalm/local-src && bun install --lockfile-only --ignore-scripts) \
+    && (cd /opt/openpalm/local-src/packages/portal-sdk && bun pm pack --destination /opt/openpalm/local-artifacts --quiet) \
+    && (cd /opt/openpalm/local-src/packages/portal-discord && bun pm pack --destination /opt/openpalm/local-artifacts --quiet) \
+    && (cd /opt/openpalm/local-src/packages/portal-slack && bun pm pack --destination /opt/openpalm/local-artifacts --quiet) \
+    && (cd /opt/openpalm/tools && bun add /opt/openpalm/local-artifacts/*.tgz --production) \
```
itlackey: **[Critical] `bun add *.tgz` does not install the local portal-sdk candidate — it resolves `@openpalm/portal-sdk` from the npm registry.**

Reproduced end-to-end with the real repo packages: after this exact recipe (`workspace.package.json` → `bun install --lockfile-only` → `bun pm pack` ×3 → `bun add /opt/openpalm/local-artifacts/*.tgz --production`), bun nests a **registry-fetched** `@openpalm/portal-sdk` under `node_modules/@openpalm/discord-portal/node_modules/` and `.../slack-portal/node_modules/` (a `diff -rq` shows those nested copies differ from the local tarball's copy at the root), and module resolution from the adapters — which is what `portal-entrypoint.ts` actually imports through — picks the nested registry copy.

Two failure modes:
1. **Release builds break**: `bump-unit.mjs` stamps a new, not-yet-published version; the adapters' packed dependency becomes `@openpalm/portal-sdk@<new>`; the registry 404s and the docker build fails. The docker job has no npm-publish dependency (`needs: [compute-version, candidate, preflight]`), so there is no ordering that saves it.
2. **Rebuilds at an already-published version** silently ship the *published* SDK instead of the local candidate — defeating the PR's stated guarantee that images bake the local source.

Fix: pin the SDK to the local tarball via an `overrides`/`resolutions` entry in `workspace.package.json` (pointing at the packed sdk tgz), or install the sdk tarball explicitly and configure bun's linker so the adapters' dependency resolves to it rather than the registry.

---

.github/workflows/release.yml
```diff
@@ -1,1445 +1,516 @@
...
-          echo "base_sha=${BASE_SHA}" >> "$GITHUB_OUTPUT"
-
-      - name: Stamp unit files
+      - name: Stamp product candidate
         env:
-          UNIT: ${{ inputs.unit }}
-          BUMP: ${{ inputs.bump }}
-          STAMP: 'true'
-          VERSION_OVERRIDE: ${{ needs.compute-version.outputs.new_version }}
+          UNIT: platform
```
itlackey, line 73: **[Major] Product releases never stamp the Electron version — shipped installers embed the previous version and self-update loops forever.**

The candidate stamps only `UNIT: platform`, which per `.github/release-package-groups.json` covers root, skeleton, lib, cli, and ui — **not** `packages/electron/package.json`. Yet the `electron` job (line 274) builds installers from this same candidate, `assemble-assets` requires them, and they ship as assets on the product GitHub Release. Nothing in the new model ever advances the electron version: `bump-unit.mjs` still has an `electron` unit, but this workflow is its only caller and it always passes `platform`.

Concrete failure: release `0.13.1` is cut while `packages/electron/package.json` still says `0.13.0-beta.13`. The release contains `OpenPalm-0.13.0-beta.13-*.zip` (electron-builder names artifacts from its own package.json); a user installs it, `app.getVersion()` returns `0.13.0-beta.13`, and `checkForElectronUpdate` sees release `0.13.1` (with installer assets) as newer — prompting an update that re-installs the same binary, endlessly. Two successive product releases also publish identically-named but different-content electron assets under different tags.

Fix: add `packages/electron/package.json` (and `admin-tools`) to the group stamped for a product release, or run a second `UNIT: electron` stamp with the same `VERSION_OVERRIDE` in this step.

---

.github/workflows/release.yml
```diff
@@ -1,1445 +1,516 @@
...
+            exit 0
           fi
+          [ "${REMOTE_SHA}" = "${BASE_SHA}" ] || { echo '::error::release branch moved'; exit 1; }
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          if git ls-remote --exit-code --tags origin "refs/tags/${VERSION}" >/dev/null 2>&1; then
+            git fetch --force origin "refs/tags/${VERSION}:refs/tags/${VERSION}"
+            test "$(git rev-parse "${VERSION}^{tree}")" = "${CANDIDATE_TREE}" || { echo "::error::Product tag ${VERSION} belongs to different source"; exit 1; }
+          else
+            git tag -a "${VERSION}" "${CANDIDATE_SHA}" -m "OpenPalm ${VERSION}"
```
itlackey, line 417: **[Major] Bare-version tags are invisible to the admin UI's releases picker.**

This tags product releases with the bare semver (`0.13.0`), but `packages/ui/src/lib/server/release-units.ts` — which feeds `GET /api/host/versions/releases` via `selectInstallableReleases` — only matches `platform-X.Y.Z`, `electron-X.Y.Z`, and legacy `vX.Y.Z` tags. Every release cut by this workflow is filtered out, so the admin UI's releases list will be permanently empty from the first new-model release onward. Meanwhile the in-app Electron update check (`normalizeVersion` accepts bare tags) *does* see them, so the two surfaces disagree.

Fix: add a bare-semver pattern to `release-units.ts` (and decide whether the unit-prefixed patterns are still reachable at all under the new model).

---

scripts/setup.sh
```diff
@@ -67,15 +64,11 @@ if [ -n "${REQUESTED_VERSION}" ]; then
...
-  if [ "${SCRIPT_VERSION}" != "main" ]; then
-    VERSION="$(normalize_version "${SCRIPT_VERSION}")"
-  else
-    LATEST_RELEASE_JSON="$(curl -fsSL "https://api.github.com/repos/itlackey/openpalm/releases/latest")" \
-      || die "Could not determine latest release version"
-    RAW_VERSION="$(printf '%s\n' "${LATEST_RELEASE_JSON}" | grep '"tag_name"' | sed -E 's/.*"tag_name": *"([^"]+)".*/\1/')"
-    [ -n "${RAW_VERSION}" ] || die "Could not determine latest release version"
-    VERSION="$(normalize_version "${RAW_VERSION}")"
-  fi
+  LATEST_RELEASE_JSON="$(curl -fsSL "https://api.github.com/repos/itlackey/openpalm/releases/latest")" \
```
itlackey: **[Major] Dropping the `SCRIPT_VERSION` stamp makes every default install depend on the unauthenticated GitHub API — and changes which version gets installed.**

Three consequences of removing the release-stamped pin (same applies to `setup.ps1`):

1. **Rate limiting**: every default install now calls `api.github.com/releases/latest` unauthenticated — 60 requests/hour/IP, with no fallback. Users behind shared NAT/CI egress IPs get HTTP 403 and `die "Could not determine latest release version"`, where the old script installed its pin with zero API calls.
2. **Prerelease semantics**: `releases/latest` excludes prereleases. During a beta-only phase, the one-liner installs the last *stable* release instead of the current beta — or 404s if no stable release exists — where the old script self-pinned to the beta it shipped with.
3. **Release-asset identity**: a copy of `setup.sh` downloaded from release X's assets no longer installs X.

If the stamp is intentionally retired, consider at least falling back to the `Location` redirect of `https://github.com/itlackey/openpalm/releases/latest` (not API-rate-limited) instead of the API call.

---

containers/portal/start.test.ts
```diff
@@ -28,22 +28,16 @@ describe('portal image bake contract', () => {
...
-  test('docker image bakes the first-party adapters from npm via tools/package.json', () => {
+  test('docker image bakes the local SDK and first-party adapter candidates', () => {
     const dockerfile = readRelative('containers/portal/Dockerfile');
-    const tools = readRelative('containers/portal/tools/package.json');
 
-    // Adapters are published npm packages installed at build time under
-    // /opt/openpalm/tools — no workspace source is copied into the image.
-    expect(dockerfile).toContain('COPY containers/portal/tools/package.json /opt/openpalm/tools/package.json');
-    expect(dockerfile).toContain('bun install --cwd /opt/openpalm/tools --production');
+    expect(dockerfile).toContain('COPY packages/portal-sdk /opt/openpalm/local-src/portal-sdk');
```
itlackey: **[Major] This assertion doesn't match the Dockerfile — the suite fails.**

The test expects `COPY packages/portal-sdk /opt/openpalm/local-src/portal-sdk`, but `containers/portal/Dockerfile:14` copies to `/opt/openpalm/local-src/packages/portal-sdk`. Verified: `bun test containers/portal/start.test.ts` → 1 fail on this branch, so CI is red.

---

packages/skeleton/system/stack/core.compose.yml
```diff
@@ -31,7 +31,7 @@
 services:
   # ── Assistant (opencode runtime — NO docker socket) ────────────────
   assistant:
-    image: ${OP_IMAGE_NAMESPACE:-openpalm}/assistant:${OP_ASSISTANT_VERSION:-latest}
+    image: ${OP_IMAGE_NAMESPACE:-openpalm}/assistant:${OP_ASSISTANT_VERSION:?OP_ASSISTANT_VERSION is required}
```
itlackey, line 34: **[Major] The new `:?` required-var interpolation breaks the fresh-clone dev flow.**

`scripts/dev-setup.sh` seeds only `OP_HOME`/`OP_PROJECT_NAME`/`OP_UI_PORT` (plus addons) into `.dev/state/stack.env` — never `OP_ASSISTANT_VERSION`/`OP_GUARDIAN_VERSION`/`OP_PORTAL_VERSION` — and `.env.example` lists them only commented out. The `image:` override in `compose.dev.yml` doesn't help because Compose interpolates each file *before* merging, and profiles don't exempt `portals.compose.yml` either (interpolation is pre-profile).

Verified: `docker compose -f core.compose.yml -f portals.compose.yml -f compose.dev.yml config` fails with `required variable OP_ASSISTANT_VERSION is missing a value` when unset, succeeds when set. So a fresh-clone `bun run dev:build` / `dev:stack` dies at compose parse. (Managed installs are fine — `ensureVersionDefaults` writes all keys.)

Fix: have `dev-setup.sh` seed the version keys, or keep a default in the interpolation (`:?` → `:-`-with-default) for the dev overlay. Same applies to `portals.compose.yml` lines 9/51/89.

---

packages/ui/src/lib/server/port-contract.ts
```diff
@@ -40,9 +42,20 @@ export function reconcileSupervisedPortContract(
...
+    ...parseEnvFile(legacyKnowledgeStackEnvFile(homeDir)),
+    ...parseEnvFile(legacyStateEnvFile(homeDir)),
+    ...readStackRuntimeEnv(homeDir),
+  };
+  const repairedDefaults = migrated.OP_ASSISTANT_PORT === '3800' && !migrated.OP_UI_PORT;
+  if (repairedDefaults) {
+    migrated.OP_ASSISTANT_PORT = '3810';
+    migrated.OP_UI_PORT = '3800';
+  }
   if (!env.OP_ASSISTANT_PORT || env.OP_ASSISTANT_PORT === '3800') {
```
itlackey: **[Major] On a fresh supervised first run this writes the literal string `"undefined"` into `process.env`, breaking the OpenCode proxy and setup wizard.**

Fresh Electron/CLI first run: `OP_UI_SUPERVISOR` is set but there's no `stack.env` and no legacy env files, so `migrated.OP_ASSISTANT_PORT`/`OP_UI_PORT` are `undefined`. The guard `!env.OP_ASSISTANT_PORT || env.OP_ASSISTANT_PORT === '3800'` then executes `process.env.OP_ASSISTANT_PORT = undefined`, which Node coerces to the string `"undefined"` (same for `OP_UI_PORT`). Reproduced by executing this function against an empty temp home.

Downstream, `opencode-target.ts:95` and `api/setup/opencode/ensure/+server.ts:116` build `http://127.0.0.1:undefined` — the `?? '3810'` fallbacks never fire because `"undefined"` is truthy — so the `/oc` assistant proxy and the setup wizard's OpenCode probe are broken for the process lifetime.

The pre-PR code couldn't hit this: it early-returned via `!runHomeMigrations(homeDir)`, which returns `false` for an absent install. Fix: skip the assignment when the migrated value is undefined (or guard the whole reconciliation on an existing install).

Related, lower severity: (a) the retired `OP_OPENCODE_URL` cleanup below was narrowed from "any migrated port move" to only the exact 3800→3810 repair, so a custom-port home with a stale generated `OP_OPENCODE_URL` no longer gets it deleted and proxies to the wrong upstream; (b) the docstring above still says this "Goes through `runHomeMigrations`", which it no longer does.

---

packages/cli/src/lib/cli-state.ts
```diff
@@ -29,14 +28,7 @@ const logger = createLogger('cli:state');
...
 export function migrateBestEffort(homeDir: string): void {
-  try {
-    runHomeMigrations(homeDir);
-  } catch (error) {
-    logger.warn('could not migrate the OP_HOME layout; continuing with what is readable', {
-      homeDir,
-      error: error instanceof Error ? error.message : String(error),
-    });
-  }
+  logger.debug('schema migration deferred to a locked lifecycle transaction', { homeDir });
```
itlackey: **[Major] Gutting `migrateBestEffort` regresses read-only commands on pre-consolidation homes — and leaves three docstrings describing behavior that no longer exists.**

`runHomeMigrations` now only runs inside `runComposeWithPreflight` (mutations) and `install`'s `prepareInstallFiles`. The calls were removed from `ensureValidState`, `resolveServeState`, and `runUiBuild` — but the docstring right below (on `ensureValidState`) still spells out exactly why that's a problem: on a schema<2 home (split `knowledge/env/stack.env` + `state/stack.state.env`; the migration still ships in `home-schema.ts`), `state/stack.env` doesn't exist, so a read-only command resolves Compose with **no env at all**.

Concrete upgrade scenario: a user on the old layout upgrades the CLI and runs `openpalm status` / `logs` / `admin` / `ui` / `app` before any mutating command. Add-ons report disabled, no addon profiles are selected, wrong project/ports — and `classifyLocalInstall` reads `OP_SETUP_COMPLETE` as unset, classifying the install as `setup_incomplete`, so the UI's setup guard bounces the operator into the setup wizard on a working install.

Also left behind:
- The docstrings on `migrateBestEffort`, `ensureValidState` ("The ONE exception is the OP_HOME schema migration, which has to run before any state is read"), `resolveServeState` ("Migrates for the same reason"), and `status.ts:31-33` are all now false.
- `uninstall.test.ts:196-205` and `reset-password.test.ts:142-151` now only assert the vacuous stub "is a function".

If deferring migration to the locked lifecycle transaction is intentional, read paths need either a read-only env fallback (read the legacy paths when `state/stack.env` is absent) or an explicit "run `openpalm install`/`update` to migrate" error — and the stale docs/tests should go. Otherwise, restore the best-effort call on the read paths.

---

packages/cli/src/commands/install.ts
```diff
@@ -7,302 +7,345 @@ import { defineAction } from '../lib/action.ts';
...
+			}
+		}
+		// #461: stop the currently-running stack BEFORE backing up OP_HOME.
+		// Backing up the home dir while the old stack is up leaves orphaned containers
+		// holding the project name + host ports, so the fresh install collides on
+		// `compose up`. Volumes are preserved (no -v). Best-effort: a Docker-down
+		// host or a never-started install simply has nothing to bring down.
+		try {
+			const existingState = createState();
+			if (existingState.artifacts.compose) {
```
itlackey, line 238: **[Major] This guard is dead code — `install --force` never stops the old stack (regresses #461).**

`createState()` (`packages/lib/src/control-plane/lifecycle.ts:82`) always returns `artifacts: { compose: '' }`; runtime artifacts are only populated by callers that assign `state.artifacts = resolveRuntimeFiles()`, which this path doesn't. So `existingState.artifacts.compose` is always `''`/falsy and `activateComposeCommand(existingState, ['down'])` never runs.

The pre-PR code checked `buildComposeOptions(existingState).files.length > 0`, which read the compose files from disk. Now a `--force` reinstall over a running stack backs up OP_HOME while the old containers keep running and holding the project name and host ports — the exact orphaned-container/port-collision bug the surrounding comment says this block prevents. No test covers this branch (nothing asserts "Stopping existing stack").

Fix: populate `existingState.artifacts = resolveRuntimeFiles()` before the check, or go back to a disk-based existence check.

---

packages/cli/src/lib/cli-compose.ts
```diff
@@ -46,73 +42,47 @@ export async function runComposeReadOnly(
...
-      // guidance, wrapped around the friendly, never-blank mapped stderr.
-      throw new Error(buildComposePreflightError(options, mapDockerError(rawStderr).message));
-    }
-  }
-
-  const composeArgs = buildComposeCliArgs(state);
-  await runComposeStreaming([...composeArgs, ...composeSubArgs], {
-    timeoutMs: composeUpTimeoutMs(),
-  });
+	await activateComposeCommand(state, composeSubArgs, {
```
itlackey, line 91: **[Major, possibly intentional] The mandatory activation audit now gates teardown commands with no escape hatch — a broken stack can become impossible to stop or uninstall.**

`runComposeActivation` (`activation.ts:44-65`) unconditionally runs `composeConfigJsonSync` + `auditFileBasedSecrets` for **every** compose mutation, including `stop`, `down`, and `uninstall --purge`. `OP_SKIP_COMPOSE_PREFLIGHT` only skips the docker-ready/disk checks above — not the activation audit.

Failure scenario: a home with a legacy secret-shaped key in `state/stack.env` (e.g. an old `OPENAI_API_KEY=` line) or a syntactically broken `custom.compose.yml` can no longer be torn down through the CLI — `openpalm stop`/`uninstall` throw `Refusing Compose down: secret-boundary audit failed` / config-resolution errors, leaving the stack running. These are exactly the installs users most need to remove. (The UI's uninstall endpoint has the same regression — see the comment there.)

Fail-closed on `up`/`apply` makes sense; for `down`/`stop`/`uninstall` consider best-effort teardown (log the audit failure, proceed), since refusing to stop containers doesn't protect any secret.

Related, from the resolved-config audit itself (`secret-audit.ts:208-217`): once any top-level secret exists in the merged project (always true — core defines two), any user service env var merely *ending* in `_FILE` (e.g. `GOOGLE_APPLICATION_CREDENTIALS_FILE: /config/gcp.json` in `custom.compose.yml`) must equal `/run/secrets/<granted>` or the whole activation refuses. That blanket suffix rule can hard-block legitimate overlays that never touched OpenPalm secrets.

---

packages/lib/src/control-plane/image-snapshots.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/** Tag old image IDs locally and persist those immutable refs for rollback. */
+export async function restoreRunningImageIds(
+  state: ControlPlaneState,
+  snapshot: RunningImageSnapshot,
+  generation: string,
+  docker: DockerClient = realDockerClient,
+): Promise<void> {
+  const env = readStackEnv(state.homeDir);
```
itlackey, line 40: **[Major] `rollback-<generation>` image pins written into `stack.env` permanently wedge every subsequent update.**

`restoreRunningImageIds` writes `OP_*_VERSION=rollback-<generation>` — tags that exist only in the local Docker daemon — into `state/stack.env`, and nothing ever resets them: `ensureVersionDefaults` only fills *missing* keys, and `performUpgrade` never rewrites pins.

Failure scenario: a deploy/upgrade over a running stack fails → rollback pins `OP_ASSISTANT_VERSION=rollback-generation-…`. Every later `openpalm update` / UI host update runs `performUpgrade` → `activateStack(..., { pull: 'always' })` → `docker compose pull` tries to fetch `openpalm/assistant:rollback-generation-…` from the registry → 404 → `pullFailed` → upgrade aborts, forever, until the operator hand-edits `stack.env`. There is no recovery path in code.

Fix: have `performUpgrade` (or the next successful update) rewrite `OP_*_VERSION` back to real release tags, or make the pull step tolerate locally-present tags (`pull_policy: missing` semantics for `rollback-*` tags).

---

packages/lib/src/control-plane/lifecycle.ts
```diff
@@ -160,188 +172,240 @@ async function reconcileCore(
...
+
+	// Migrate BEFORE snapshotting: on a pre-consolidation home the snapshot
+	// list's state/stack.env does not exist yet, so snapshotting first would
+	// capture no stack env at all and a failed deploy could not roll back env
+	// mutations. The migration is value-preserving, so the snapshot still
+	// records the pre-update values — just in the canonical location.
+	// Migrations are destructive lifecycle work. The complete safety backup is
+	// deliberately created while the caller's install lock is held and before
+	// the first migration can write anything.
+	backupOpenPalmHome(state.homeDir);
```
itlackey: **[Major] Full OP_HOME backup on every apply, with no retention at this call site and a fail-closed space gate.**

Every `applyInstall`/`applyUpdate`/`performUpgrade`/`runDeploy` now copies the entire home — everything except `data/` and `cache/`, which includes `workspace/` and `knowledge/`, potentially GBs of user files — into `data/backups/<timestamp>`. This call site never prunes: `pruneBackupDirs` is only invoked by the CLI `install` command for its *own* separate backup (which also means CLI install now takes two full back-to-back backups). And `backupOpenPalmHome` **throws** when `checkBackupFreeSpace` deems space insufficient.

Failure scenario: a user with a 5 GB workspace toggles addons / runs updates a few dozen times → dozens of duplicated home copies → disk fills → every subsequent lifecycle operation hard-fails with a backup-space error. The disk-full state is self-inflicted and self-perpetuating.

Fix: prune (`pruneBackupDirs(homeDir, N)`) right after a successful backup here, and consider excluding `workspace/`/`knowledge/` (or making the pre-apply backup scope just the managed files this apply will touch).

---

packages/lib/src/control-plane/ui-assets.ts
```diff
@@ -1,781 +1,170 @@
...
-  } finally {
+    const selectedChannel = uiUpdateChannel(platformVersion, channel);
+    const { staging, manifest } = await stageUi(toHostChannel(selectedChannel) === 'prerelease' ? 'next' : 'latest', dataDir, selectedChannel);
+    if (typeof harnessContract === 'number' && manifest.minHarnessContract > harnessContract) {
+      rmSync(staging, { recursive: true, force: true });
+      return { updated: false, latestVersion: manifest.platformVersion, redownloadRequired: true, requiredHarnessContract: manifest.minHarnessContract };
+    }
+    const current = readUiBuildVersion(resolveUiBuildDir());
+    if (current && compareComparableVersions(manifest.platformVersion, current) <= 0) { rmSync(staging, { recursive: true, force: true }); return { updated: false, latestVersion: manifest.platformVersion }; }
+    if (existsSync(uiDir)) { backup = join(resolveBackupsDir(), `ui-${Date.now()}`); mkdirSync(resolveBackupsDir(), { recursive: true }); renameSync(uiDir, backup); }
```
itlackey, line 153: **[Major] `ui-*`/`skeleton-*` backup retention regressed — `pruneBackupNamespace` lost its only production caller.**

The deleted `npm-bundle-updater.ts` pruned its own `ui-<ts>`/`skeleton-<ts>` backup namespaces after each successful hot-swap (`pruneBackupNamespace(...)`, added precisely because — quoting its old comment — nothing else pruned them and they accumulated indefinitely). The new `checkAndUpdateUiBuild`/`checkAndUpdateSkeleton` create the same backups here and at line 162 but never prune. `pruneBackupNamespace` is now exported dead code (only `index.ts` and tests reference it).

Failure scenario: a host on the prerelease channel updates frequently → one full UI-build copy plus one system-tree copy per update accumulate in `data/backups/` forever — the exact bug the removed code documented and fixed. (Namespace-aware pruning does run on `openpalm install`, but that's rare relative to hot-swap updates.)

Also: the docstrings updated by this PR now point at the wrong writer — `backup.ts:368` and `home.ts:238` say these prefixes are "written by host-assets-updater.ts", but `ui-assets.ts` writes them (and `pruneBackupNamespace`'s own docstring, "Used by the host-side hot-swap updater", is no longer true of any caller).

Fix: call `pruneBackupNamespace(homeDir, 'ui'|'skeleton', N)` after each successful swap, and correct the comments.

---

packages/lib/src/control-plane/ui-assets.ts
```diff
@@ -1,781 +1,170 @@
...
- * Update path: download the skeleton and atomically replace OP_HOME/system/ from
- * the staged tree's system/ subtree. Non-destructive — the staged copy is
- * validated before the live system/ is touched.
- */
-async function downloadNpmSkeletonBundle(manifest: NpmSkeletonManifest, homeDir: string, dataDir: string): Promise<void> {
-  const staging = await stageSkeletonDownload(manifest, dataDir);
+export interface UiBuildUpdateResult { updated: boolean; latestVersion: string | null; error?: string; redownloadRequired?: boolean; requiredHarnessContract?: number; backupDir?: string; }
+export async function checkAndUpdateUiBuild(platformVersion: string, dataDir: string, channel?: UiUpdateChannel, harnessContract?: number | null): Promise<UiBuildUpdateResult> {
+  const uiDir = join(dataDir, 'ui');
+  let backup: string | undefined;
```
itlackey, line 142: **[Major] The major-version auto-update guard was dropped from the UI/skeleton hot-swap.**

The old updater refused to auto-cross a major version (`isSameMajorVersion` policy — "Never auto-crosses a major version"). The rewritten `checkAndUpdateUiBuild`/`checkAndUpdateSkeleton` only check `compareComparableVersions(release, current) <= 0`. The skeleton path has no gate at all, and the UI path's `minHarnessContract` gate is skipped for CLI callers (`packages/cli/src/commands/update.ts:41` passes no `harnessContract`).

Failure scenario: the project ships `1.0.0` → a `0.13.x` host's background check on next start silently swaps the 1.0 system tree and UI in over a 0.13 lib and running stack — no operator opt-in, no contract check on the CLI path.

Fix: restore the same-major gate (or make `minHarnessContract` mandatory on all callers) before the swap.

---

packages/lib/src/control-plane/versions.ts
```diff
@@ -34,24 +34,24 @@ const VERSION_KEY_SET: ReadonlySet<string> = new Set(SERVICE_VERSION_KEYS);
 
 /** Default values seeded into a fresh stack.env (and returned for unset keys). */
 export const VERSION_DEFAULTS: Record<VersionKey, string> = {
-  OP_ASSISTANT_VERSION: "latest",
-  OP_GUARDIAN_VERSION: "latest",
-  OP_PORTAL_VERSION: "latest",
-  OP_VOICE_VERSION: "latest",
+	OP_ASSISTANT_VERSION: PLATFORM_VERSION,
+	OP_GUARDIAN_VERSION: PLATFORM_VERSION,
```
itlackey, line 44: **[Major, medium confidence] Install-time exact image pins are never advanced by any upgrade path — updates ship new compose files with permanently old images.**

Fresh installs write `OP_ASSISTANT/GUARDIAN/PORTAL_VERSION=<install-time PLATFORM_VERSION>` into `stack.env` (here and `setup.ts:419`). But `performUpgrade`, `applyInstall`, `runDeploy`, and the host-assets updaters never rewrite these keys — `ensureVersionDefaults` fills only *missing* keys — and the new skeleton compose *requires* them (`:?`).

So `openpalm update` / UI "update" refreshes the system tree, skeleton, and UI to the new release but keeps pulling the old container images forever. Failure scenario: install at 0.13.0 → release 0.13.5 changes an entrypoint contract in both compose and image → update writes the 0.13.5 compose files, pulls the still-pinned 0.13.0 images, and boots a mismatched stack — or simply never delivers container fixes.

I could not find any pin-advancing code path in the repo. If the intent is "operator advances image versions manually via the UI", that needs documenting (and `update`'s help text adjusted); otherwise `performUpgrade` should rewrite the pins to the target release.

---

packages/skeleton/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@openpalm/skeleton",
   "version": "0.13.0-beta.13",
   "description": "OpenPalm default config skeleton — seeded into OP_HOME on install",
-  "private": false,
+  "private": true,
```
itlackey, line 5: **[Major, latent] Making `@openpalm/skeleton` private permanently freezes the upgrade-path smoke's era pool.**

`scripts/upgrade-path-smoke.sh` and `scripts/rootless-smoke-fixture.sh:40` seed old-era homes via `npm pack "@openpalm/skeleton@<version>"`, and their comments (plus `ci.yml:190-196`) describe a model where "each release publishes another era". With skeleton now private, that premise is gone. CI still passes today only because the pinned eras (`0.12.43`, `0.13.0-beta.13`) were published under the old model and npm keeps them immutably — but no release cut after this PR can ever be seeded, so upgrade-from coverage is frozen at pre-PR eras forever.

Suggest migrating era-seeding to the new GitHub host-assets archive (`openpalm-host-assets-<version>.tar.gz`), or at least documenting the frozen pool in the smoke scripts so the drift is deliberate.

---

packages/ui/src/routes/api/host/uninstall/+server.ts
```diff
@@ -43,7 +42,7 @@ export const POST: RequestHandler = async (event) => {
             requestId,
           );
         }
-        await composeDown(buildComposeOptions(state));
+        await activateComposeCommand(state, ['down'], { lock });
```
itlackey: **[Minor] Uninstall now fails closed on stacks that are too broken to pass activation — the installs users most want to remove.**

The old code called `composeDown` and ignored its failure (best-effort teardown). `activateComposeCommand` runs the mandatory config-resolution + secret-boundary audit (`activation.ts:40-66`) and **throws**, so an unresolvable compose config or an audit-failing `stack.env` now returns a 500 `uninstall_failed` *before* `applyUninstall` runs. Suggest catching activation failures here and proceeding with `applyUninstall` anyway (logging the teardown failure), mirroring the previous best-effort semantics. Same theme as the CLI-side comment on `cli-compose.ts`.

---

itlackey (commented):
# Review: deployment-model refactor

**Scope**: the exact PR diff `92570e8..846ec8e` (311 files, +8,619/−12,190), reviewed on a checkout of `refactor/deployment-model` at the head SHA. Findings were produced by parallel review passes over control-plane, containers, CI/release, CLI, UI, package moves, and a repo-wide stale-reference sweep, then each candidate was verified against the code (several reproduced by execution) before being reported. Findings from files/lines not touched by this PR were discarded.

## Overview

This PR meaningfully simplifies the deployment model: the UI is packed and baked into the assistant image from local source, the skeleton bake and runtime npm installs are gone, ui/skeleton/lib become private workspaces, portal adapters move under `packages/`, extensions get their own publish workflow, and release.yml shrinks from ~1,050 to ~520 lines. The direction is good and most of the refactor is remarkably self-consistent — the rename carried nearly all its callers, the lockfile/workspace config are clean, `bun test` passes for the portal/guardian/UI-server suites, `svelte-check` is clean, and no auth guards were lost on host endpoints.

That said, the review found **3 critical, ~10 major, and a tail of minor issues**, detailed as inline comments. The three criticals are all in the new build/publish paths and are deterministic (two reproduced by running the actual commands): the assistant image build fails at the UI pack step, the portal image installs the registry portal-sdk instead of the local candidate, and the new publish-extensions workflow fails every run on a heredoc indentation bug. A recurring theme in the majors: several load-bearing behaviors were deleted with the old model (namespace backup pruning, schema migration on read paths, the major-version update gate, the `SCRIPT_VERSION` pin, electron stamping) without their replacement actually being wired up.

## Inline comments index

| Severity | Where | Issue |
|---|---|---|
| Critical | `containers/assistant/Dockerfile` | UI pack step fails — workspace dep unresolvable (reproduced) |
| Critical | `containers/portal/Dockerfile` | `bun add *.tgz` resolves portal-sdk from registry, not local candidate (reproduced) |
| Critical | `.github/workflows/publish-extensions.yml` | Unterminated heredoc — workflow fails every dispatch (reproduced) |
| Major | `.github/workflows/release.yml` (stamp step) | Electron version never stamped → self-update loop on shipped installers |
| Major | `.github/workflows/release.yml` (tag step) | Bare tags invisible to admin UI releases picker |
| Major | `containers/portal/start.test.ts` | Assertion doesn't match Dockerfile — suite fails (CI red) |
| Major | `packages/skeleton/system/stack/core.compose.yml` | `:?` required vars break fresh-clone dev flow (verified) |
| Major | `packages/ui/src/lib/server/port-contract.ts` | Writes literal `"undefined"` ports into `process.env` on first run (reproduced) |
| Major | `packages/cli/src/lib/cli-state.ts` | Migration removed from read paths; stale docstrings; setup-wizard bounce on old homes |
| Major | `packages/cli/src/commands/install.ts` | Dead guard — `--force` never stops the old stack (#461 regression) |
| Major | `packages/cli/src/lib/cli-compose.ts` | Activation audit gates `stop`/`down`/`uninstall` with no escape hatch |
| Major | `packages/lib/src/control-plane/image-snapshots.ts` | `rollback-*` pins permanently wedge later updates (pull 404) |
| Major | `packages/lib/src/control-plane/lifecycle.ts` | Unbounded full-home backups on every apply + fail-closed space gate |
| Major | `packages/lib/src/control-plane/ui-assets.ts` | `ui-*`/`skeleton-*` backup pruning regressed (dead `pruneBackupNamespace`) |
| Major | `packages/lib/src/control-plane/ui-assets.ts` | Major-version auto-update guard removed from hot-swap |
| Major | `packages/lib/src/control-plane/versions.ts` | Image pins never advanced by updates → new compose, old images |
| Major | `packages/skeleton/package.json` | Private skeleton freezes upgrade-path smoke's era pool |
| Major | `scripts/setup.sh` | Lost version pin → API rate limits, wrong version during betas |
| Minor | `packages/ui/.../uninstall/+server.ts` | Uninstall fails closed on broken stacks |

## Additional minor findings (no inline comment)

**Correctness / robustness**
- `packages/lib/src/control-plane/host-assets-updater.ts:73` — in the listing path, `releaseVersion(release)` is called *outside* the per-release `try`, so one non-draft release with an unparsable tag makes `resolveHostAssetsRelease` throw for everyone. Move it inside the try/catch. Also (line 91): the fixed `.host-assets.staging` path means concurrent update checks (ui-server startup vs. user-triggered update, outside the install lock) can `rmSync` each other's staging mid-extract.
- `packages/lib/src/control-plane/deploy.ts:474-481` — when the *resolved-config* audit inside `activateStack` fails (which the per-overlay `auditApplyState` can't catch), the error is rethrown without `restoreDeployFiles(state)`: the refused config stays on disk and the UI gets an exception instead of a structured `deployError`, unlike the pre-activation audit path.
- `packages/lib/src/control-plane/rollback.ts` — three smaller items: generation dirs under `data/rollback/` are never pruned (line 59-91); `restoreSnapshot` now deletes live files absent from the snapshot, which on an early-install snapshot can delete a just-bootstrapped `state/stack.env` (line 129-133); and pre-PR flat-layout snapshots are invisible to the new `hasSnapshot()` right after upgrade (no `MIGRATIONS` entry converts them). In `packages/cli/src/commands/rollback.ts:96-105`, the recovery path re-ups with `managedServices` computed from the rolled-back target config (not the pre-rollback generation), and a recovery throw masks the root-cause error.
- `.github/workflows/publish-extensions.yml` — the 4-way exact-version lockstep it enforces conflicts with `release-package-groups.json`/CI treating guardian and portals as independently-versioned units, and no tool stamps all four together: after a legitimate guardian-only bump the workflow can never validate. Also: line 64 interpolates `${{ inputs.version }}` directly into bash (everywhere else the PR routes versions through `env:`), and unlike release.yml's publish job it skips the `npm install -g npm@latest` OIDC hardening.
- `scripts/build-host-assets.mjs` — reads `argv[2] ?? root package.json version` but release.yml passes `env: VERSION` and no argv; works only while the stamped root version equals the dispatched version. Dead plumbing worth aligning.
- `packages/ui/src/routes/api/host/versions/ui/+server.ts` — rewritten to GitHub host-assets, but repo-wide grep finds zero consumers of this route and it has no tests; consider deleting instead of carrying it.

**Test-quality (this one bit the PR directly)**
- `scripts/release-publish-dag.test.ts` went from a 928-line behavioral suite to ~100 lines of `toContain()` source-substring assertions; the deleted 977-line `ui-assets.test.ts` similarly became string-grep checks (`host-assets-updater.test.ts`), leaving the download/checksum/extract/swap paths with zero functional coverage (`fetch` isn't injectable as written). The always-failing publish-extensions workflow sails through the suite — a behavioral check on the embedded scripts would have caught it.

**Stale docs/comments left behind by the refactor**
- `.github/CONTRIBUTING.md:9` still lists `portals/*` in the repo layout (file otherwise updated by this PR).
- `.dockerignore:25` — dead `!portals/**/*.md` negation; also, the newly COPY'd trees have no exclusions for `packages/ui/.svelte-kit` (created by release's `ui:build` right before the docker build) or per-package `node_modules`, which land in a permanent COPY layer (the later `rm -rf` is a separate RUN → whiteout only) and cache-bust it.
- `packages/skeleton/system/stack/core.compose.yml:188-190` — healthcheck rationale cites a boot-time `npm install @openpalm/ui` that no longer exists.
- `packages/ui/vite.config.ts:63` — "LOAD-BEARING for the npm publish" rationale is obsolete (`@openpalm/ui` is private); the constraint still holds, but for host-assets/image delivery.
- `packages/ui/src/lib/server/features.ts:116` — dangling comment fragment from the `skeletonVersion` removal.
- `packages/lib/src/control-plane/core-assets.test.ts:6` — points to the deleted `ui-assets.test.ts` for `applyHomeSeed` coverage that no longer exists.
- `.github/workflows/ci.yml:278-279` — stale comment + `ROOT_VERSION` computed but unused after the setup-script version check removal.
- `.github/workflows/publish-npm-package.yml` — orphaned by this PR (zero callers; `release-publish-dag.test.ts` even asserts release.yml doesn't reference it) yet still capable of publishing the now-private packages if re-wired; suggest deleting.
- Release-workflow duplication worth consolidating: the tag-tree-equals-candidate bash blob appears twice (preflight + release-source), and the six CLI asset names are enumerated in six places (release.yml matrix, publish-bootstrap loop, `validate-release-assets.mjs`, `setup.sh`, `setup.ps1`, `self-update.ts`).

## Verified clean (checked, not assumed)

- Portal/guardian package moves: only `repository.directory` changed; `bun pm pack` resolves `workspace:*` to concrete versions; tarball names match the publish globs; `bun install --frozen-lockfile` and the workflow's exact test command pass (561 tests).
- No live references remain to `ui-kit`, `Dockerfile.local-packages`, `npm-bundle-updater`, `verify-release-images`, `bump-platform`, or `containers/portal/tools` (historical roadmap docs only).
- All `@openpalm/lib` imports across cli/electron/ui resolve against the new `index.ts`; `tsc --noEmit` clean in cli and electron; ui-kit merge is complete with genuinely equivalent replacement tests; no auth regressions on touched host endpoints.
- release.yml's own heredocs terminate correctly; checksum generation/validation is fail-closed; CLI asset names cross-check between release.yml, `validate-release-assets.mjs`, and the setup scripts; guardian's pack path has no workspace deps; both final images run non-root; no secrets baked.

---

PR #598, opened 2026-08-04
fix(ui): move sign out into settings and only offer it when it works

itlackey:
The sign-out control was a button absolutely positioned over the chat
content area, floating above the conversation on every visit. It was also
rendered unconditionally, which made it actively harmful in the
client-only lane: a process with no local install and no login password
has no login wall at all, so signing out POSTed a logout and then landed
the user on a /login whose own POST answers 503 admin_not_configured.
There was no way back into the app from there.

It now lives in Settings -> General as a Session section, gated on a new
`signedIn` field from a page-scoped server load. That field mirrors
`event.locals.role`, which hooks.server.ts resolves from
identifyCallerByToken() -- and which is null whenever no password is
configured. So the control appears exactly when there is a session to
end and a login page able to take the user back, and is absent
everywhere else.

The fact is page-scoped rather than added to the root layout payload,
which is the RuntimeContext v2 contract and has no business carrying a
single page's concern.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01XAXhuEE252G214x4KBnUsb


itlackey:
Quality gates went red on `b785fc8`, not on the UX commits — the run before it (`4e088ab`) was fully green. Since that commit is yours, flagging what I found and what I changed rather than quietly rewriting it.

**Root cause.** `ORIGIN: ${OP_UI_ORIGIN:-}` sets the variable to an *empty string* when `OP_UI_ORIGIN` is unset, and `OP_UI_ORIGIN` is defined nowhere in the tree — so this is the default path for every deployment, not an edge case. adapter-node's `parse_origin` accepts a valid absolute URL or an **absent** variable and throws on everything else, empty string included:

```js
if (value === undefined) return undefined;
const trimmed = value.trim();
try { url = new URL(trimmed); } catch { throw new Error(`Invalid ORIGIN: '${trimmed}'...`) }
```

It runs at module load, so the UI child died before it could listen. The assistant healthcheck curls `http://localhost:3000/health`, that failed, the container never went healthy, and compose gave up with `dependency failed to start` — which is exactly the timeout in the log.

Reproduced against a real adapter-node build rather than by inspection:

```
$ ORIGIN='' node build/index.js
Error: Invalid ORIGIN: ''. ORIGIN must be a valid URL with http:// or https:// protocol.
```

**Fix** (`445cb07`): `start_ui` now drops `ORIGIN` from the child's env when it is empty. Your knob is untouched — set `OP_UI_ORIGIN` and it is forwarded as before. The same `/health` probe the healthcheck runs returns 200 for empty, unset, and pinned values alike, and there's a regression test covering all three plus the compose/entrypoint pairing.

**One thing worth your call.** The reason you added this is real: adapter-node defaults the protocol to `https` when no `x-forwarded-proto` arrives, so a direct plain-HTTP LAN install computes an `https://…` origin. My fix stops the crash but doesn't change that default — with `OP_UI_ORIGIN` unset you still get the https assumption you were trying to correct. If you want it fixed by default rather than per-operator, that's a design question (derive the origin from `OP_UI_BIND_ADDRESS`/`OP_UI_PORT`? always send `x-forwarded-proto`?) and I'd rather you pick the direction than guess.

---

PR #564, opened 2026-07-12
0.13.0 milestone epic

fwdslsh-dev:
## Summary

- Completes the shippable OpenPalm 0.13.0 milestone scope across the standalone client, remote connections and pairing, portal packaging, network presets, mDNS discovery, Guardian state/CORS hardening, and release automation.
- Removes the `channel_lan` / `CHANNEL_NAME` compatibility surfaces and adds upgrade-time protection for stale custom overlays.
- Finishes localhost PWA installability, runtime compatibility checks, encrypted connection storage, credential-safe offline behavior, and client-side HTTPS enforcement for hosted origins.
- Qualifies the release through the full local test matrix, source-stack smoke tests, PR CI, package/image packing, and a coordinated beta release dry run.

## Issue Closure

Closes #433
Closes #486
Closes #488
Closes #490
Closes #491
Closes #511
Closes #557
Closes #563

#435 and #506 intentionally remain open; their remaining work is documented on the issues.

Scope decisions recorded in this branch:

- #433: Phase 2 ownership/A2A persistence was declined under the issue's optional evidence gate; unused registry columns remain ALTER-on-consumer. WAL, versioned migrations, private file modes, and complete principal API CRUD shipped.
- #511: the repository-controlled PWA scope shipped. The official hosted-origin deploy and matching default CORS entry were explicitly moved to 0.14+ because no hosting provider, DNS, or deployment credentials exist yet.
- #557: TLS termination is an infrastructure concern. The supported Tailscale/Caddy paths, exact-origin CORS, HTTPS refusal, and operator configuration are implemented and documented; no custom Guardian TLS relay ships.

## Verification

- All six PR checks pass at `8ef885b8`: Stack hygiene, Biome lint, Quality gates, UI unit tests, Electron unit tests, and Client browser tests.
- Root non-UI aggregate: 2,001 passed, 0 failed.
- UI Vitest: 1,599 passed.
- Guardian: 260 passed.
- Electron: 134 passed.
- Client Playwright: 16 passed; mocked UI E2E: 8 passed; live source-stack UI E2E: 73 passed.
- Docker install/update invariants, rootless source-stack smoke, portal smoke, package packing, CLI builds, and PWA server-down/offline recovery checks passed.
- Coordinated `0.13.0-beta.4` `unit=all` dry run passed without publishing packages, images, tags, or a release: https://github.com/itlackey/openpalm/actions/runs/29301627841

PR #627, opened 2026-08-16
One credential for OpenCode, and /advanced stops faking a chat surface

itlackey:
Two changes, both from the original ask: `/advanced` should show OpenCode, and there should be one credential rather than two.

Net effect on non-test code is **−105 lines** (301 in / 406 out) against `73a2f62`.

## 1. `/advanced` no longer renders a fake chat surface

When the workspace was unreachable it rendered a partial duplicate of `/chat` — a composer, a message list, permission cards — under a heading promising OpenCode. Every deployment problem looked like a broken app, and the route became a worse copy of the surface that already worked.

It now says the workspace is unavailable and links to `/chat`. There is deliberately **no** chat fallback; a test pins that it never renders a second composer.

## 2. OpenCode always requires a password — `OPENCODE_AUTH` deleted

`OPENCODE_AUTH` tracked `assistantDirect`, so OpenCode authenticated only while its port was published. Three consequences:

- the **default install ran OpenCode with no password at all**;
- turning the toggle off silently removed the credential from a running server;
- the assistant entrypoint, the guardian, two healthchecks, a scheduled task and the host resolver each carried their own copy of the gate — disagreement produced a 401 storm rather than an error anyone could see.

One answer now: the `op_opencode_password` secret, which `ensureSecrets` always materializes. `assistantDirect` becomes what its label always claimed — a bind, deciding who can reach a password-protected port.

| | |
|---|---|
| lib | `resolveOpenCodeCredential` unconditional, and no longer takes a `persistedEnv` (the stack env is not an input to a credential); `resolveAccessEnv` stops writing the key; `requiresAssistantKey` deleted; setup mints the key on every install |
| entrypoint | resolves the secret unconditionally; `opencode_auth_enabled`, `is_loopback_address` and the LAN-exposure warning go with the posture they described |
| compose | both `OPENCODE_AUTH` rows and both healthcheck case-statements gone; the probes authenticate unconditionally |
| guardian | always attaches upstream auth |
| migration | `HOME_SCHEMA_VERSION` 8 → 9 sweeps the stale row |
| fixtures | `dev-setup.sh` and the rootless smoke fixture seed real values — "empty file = auth off" is no longer a posture |
| copy | the connections page claimed the password was "only required" with a toggle on, which was false for an OpenPalm assistant |

## The workspace address

`/advanced` opens the page's own host and scheme plus `OP_WORKSPACE_PORT`. One number, no override.

The session cookie is what authenticates the workspace and cookies are scoped by **host**, so the workspace lives on the UI's hostname. Cookies ignore **ports**, so a second port is free — which covers every layout that works: desktop, LAN, Tailscale's second serve entry, Caddy's second port block.

Cloudflare Tunnel cannot host the embedded workspace: it proxies a fixed port set that `OP_WORKSPACE_PORT` is not in. That is now documented — Cloudflare previously was not mentioned at all. `/chat` works there fully.

## Verification

Real Chromium against the built UI, on a **default** install (nothing published, no toggle set), with an upstream that refuses every unauthenticated request:

```
PASS  derived port on the page's own host (no configuration)
      iframe #root: "FAKE-OPENCODE-ROOT API-OK"   after reload: same
```

The assertion is made *inside* the iframe — the upstream's marker plus the result of a root-absolute `/api` call the frame made for itself. So OpenCode is password-protected out of the box, and the OpenPalm session alone reaches it.

Tailscale's serve entry was verified in code (`endpointsFor` writes `${TS_CERT_DOMAIN}:3820` beside `:443`, non-funnelable), not against a live tailnet.

Suites: lib 1577, cli 230, guardian 220 (including the `config.test.ts` that was missing), UI server 1758, advanced client 4 — all pass. `ui:check` clean. The root `bun run test` run sits at its pre-existing 149-failure Docker-dependent baseline, and lint at its one pre-existing `install.ts` warning; both confirmed unchanged by stashing the diff.

## Upgrade notes

- **Schema 8 → 9 is one-way.** Older binaries are refused on write paths — deliberate, since their resolver honours `OPENCODE_AUTH` and would attach no credential against always-authenticated containers.
- **Host-local unauthenticated clients break.** `127.0.0.1:3810` had no password by default before; anything hitting it directly now gets 401 until it picks up the key from `private/secrets/op_opencode_password` (or the dashboard). That break *is* the fix.

## History

The branch contains three revert commits. The ticket-based cross-hostname handoff (`15c6ac3` → `92dcd26`), the boot guards added around always-on auth (`a2ba621` → `5bd2fb6`), and a cookie-domain setting (`9d819d8` → `fffa446`) were each built and then removed as out of scope. They are left in the history rather than squashed away, since the reverts explain why the surviving design is as small as it is.

https://claude.ai/code/session_01WiD3ivt4SjdTau3vjoYhfx

---
_Generated by [Claude Code](https://claude.ai/code/session_01WiD3ivt4SjdTau3vjoYhfx)_


itlackey:
Thanks — three of the six were real defects and are fixed in e914ce5. The other three I'm declining, with reasoning.

### Fixed

- **`docs/remote-access-tls.md`** — the stray `setting the cookie domain.` fragment and the duplicated Caddy tail (DNS/CORS/connection URL) under Cloudflare Tunnel. Both were mine, from an imprecise slice when an earlier setting was removed. The Caddy-specific firewall paragraph had been dragged along with it and is moved back under Caddy.
- **`features.vitest.ts`** — the stray blank lines a deleted `describe` left behind.

### Declining: fail closed on a blank/missing password

Three comments (`entrypoint.sh:131`, `guardian/config.ts:131`, `secrets-files.test.ts:129`) ask for a boot-time failure when the password cannot be resolved. That behaviour existed earlier in this branch and was removed deliberately. Restoring it would reintroduce the specific failure it caused:

`ensureSecret` treated a **0-byte** file as torn and re-seeded it, but a file containing one newline is **1 byte**. It passed that check *and* the entrypoint's `[ -s ]` guard, then read back empty once `$(cat)` stripped the trailing newline. With fail-closed in place, a single stray newline in `op_opencode_password` — from a hand edit or a restored backup — took down the assistant and the guardian at boot.

The fix attempted for that was worse: `ensureSecret` silently regenerating a non-empty file the operator had put there. A guard on a guard, with the second one overwriting operator-owned state.

Two specifics on the guardian comment:

- Returning `null` is not a 401 storm, it is the **matching** behaviour. The guardian and the assistant read the same file the same way, so "no password here" means "no password there" — the assistant is serving unauthenticated and sending no header is correct. Throwing instead produces a guardian that refuses to boot against an assistant that is running fine, which is the dead-stack-that-looks-healthy failure inverted.
- Unconditional resolution at module load also made `config.ts` unimportable by any tool or test without the compose secret mounted. Working around that needed a memoized accessor, an explicit boot assert, a `bunfig.toml` and a test preload in two places — all of it servicing a throw that shouldn't have been there.

The design point: OpenCode is authenticated **by default**, not by enforcement. Setup generates the key on every install, Compose always grants it, Guardian sends the same value — so the default install is authenticated and publication never changes that. An operator who empties their own secret gets an unauthenticated OpenCode and a startup warning, because the file is theirs.

### The real finding underneath those three

The invariant Copilot cites (`core-principles.md:49`, "requires Basic auth in EVERY configuration") **was** inconsistent with the code — that wording survived the removal of the fail-closed guard in fourteen places. That is a genuine contradiction and it is now fixed: the docs and comments say "authenticated by default", and spell out that emptying the secret disables it. Same commit.

---

itlackey:
@codex check this PR and do a sanity check on the solution design

itlackey:
Both Codex findings were real. Fixed in 1f89549.

They're the same underlying thing: the design here is "blank secret = no auth", and it was implemented inconsistently in two places.

### 1. `guardian/config.ts` — the guardian disagreed with the other two readers

Confirmed by tracing all three:

| reader | `"   \n"` becomes | verdict |
|---|---|---|
| `entrypoint.sh` | `$(cat)` → `"   "`, then `[ -z ]` | password |
| `lib` `resolveOpenCodeCredential` | `stripTrailingNewlines` → `"   "`, truthy | password |
| `guardian` (before) | `.replace(/\n+$/)` → `"   "`, then `.trim() === ""` | **no password** |

So OpenCode started with a three-space password, the UI proxy sent three spaces, and the guardian alone sent nothing — every portal request 401ing against an assistant that direct UI traffic reached fine. Exactly the trim-divergence class as #564, in a new place. Now `password === ""` is the only "no credential" case and all three agree.

### 2. `setup.ts` — a rerun regenerated a deliberately blanked key

Sharper than the report, once I checked the ordering: `ensureSecrets` seeds `op_opencode_password` unconditionally at line 399, **before** this branch at 447. So by the time the branch runs the file always exists non-empty — *unless* an operator emptied it on purpose. The branch could only ever fire in the one case where firing was wrong.

Deleted rather than guarded. The seeding its comment claimed to do was already happening above it.

### Pinned

- guardian suite: whitespace **is** a password; only exactly-empty is not
- `setup.test.ts`: a rerun preserves a blanked key

lib 1578 · cli 230 · guardian 221 · UI server 1758 — all pass. `ui:check` clean.

Worth noting these are the opposite kind of finding from the earlier Copilot review, which asked to fail closed on a blank secret. That would have removed the posture these two bugs were mishandling; these fix the mishandling and leave the posture alone.

---

PR #647, opened 2026-09-02
fix: 0.13.1 bugfix set (#645 #643 #641 #642 #639 #636 #635 #644 #632 #633)

### LINKED ISSUE #632: install --no-start leaves a home that cannot docker compose up
Found during 0.13.0 fresh-install verification (\`openpalm install --no-start --yes\` into an isolated \`OP_HOME\`).

\`op_ui_login_password\` is never generated by \`install --no-start\` — it's set by the wizard, which this flag deliberately skips. The first \`docker compose up\` against that home then fails opaquely:

\`\`\`
Error response from daemon: invalid mount config for type "bind": bind source path does not exist: <OP_HOME>/state/secrets/op_ui_login_password
\`\`\`

Nothing in the CLI output or the compose error explains why, or what to do about it. An operator (or a script) using \`--no-start\` for any reason other than immediately opening the wizard hits this with no signal pointing at the cause.

## Fix options, least to most invasive

1. At minimum: a clearer error at \`compose up\` time (or a preflight check in the CLI) naming the missing secret and pointing at the wizard / \`openpalm start\`.
2. Better: \`install --no-start\` could seed a placeholder secret and have the wizard rotate it on first real login, so a home is always compose-able immediately after install.
3. Document the constraint explicitly wherever \`--no-start\` is described, if the intent is that it's genuinely only for wizard-pending states.

Not a regression in 0.13.0 — this behavior appears pre-existing — but never previously verified because the fresh-install lane hadn't been exercised end-to-end before this cycle.

### LINKED ISSUE #633: bun.lock drifts from platform package.json versions across release cycles
\`bun.lock\` still records \`0.13.0-beta.31\` for the platform workspaces (5 occurrences) while every \`package.json\` in the platform tree reads \`0.13.0-beta.33\` / \`0.13.0\`. Last lockfile touch was the beta.31 prep commit; subsequent version bumps didn't refresh it.

## Why it matters

A stale lockfile self-referential version doesn't change what gets installed (workspace versions are resolved from \`package.json\`, not the lock's own version field), so this hasn't caused a functional break. But it's a release-hygiene gap worth closing:
- It's confusing signal when debugging — \`bun.lock\` disagreeing with every manifest looks like a real problem until you check closely.
- It suggests the release process doesn't run \`bun install\` (or equivalent) as part of the version-bump step, which means the lockfile could drift further in ways that DO matter (a dependency range resolving differently) without anyone noticing.

## Fix

Confirm whether \`.github/workflows/release.yml\`'s candidate-preparation step runs \`bun install\` after bumping versions — if not, add it. If it does and the lockfile still isn't updated, that's a bug in the bump step worth finding. Either way, run \`bun install\` once now to true up the lockfile as part of closing this issue.

### LINKED ISSUE #635: Desktop app cannot complete its self-update: quitAndInstallCalled jams the updater and the app reverts to the old version
Reported on both Linux and macOS. Evidence below is from a Linux AppImage install, `~/.config/@openpalm/electron/logs/main.log`.

## Symptom

The app updates, runs the new version once, then comes back on the old one and offers the same update again.

```
2026-08-31T16:31:59.311Z [warn] install call ignored: quitAndInstallCalled is set to true
2026-08-31T16:32:00.651Z [info] OpenPalm starting (v0.13.0)
...
2026-08-31T17:38:11.024Z [info] OpenPalm starting (v0.12.42)
2026-08-31T17:38:12.544Z [info] App update available: v0.13.0
```

Across the whole log: **24 starts at v0.12.42, exactly 1 at v0.13.0.**

`install call ignored: quitAndInstallCalled is set to true` occurs 5 times over several weeks (2026-08-04, 08-11 x3, 08-31). That is electron-updater's guard: once `quitAndInstall()` has been called, later calls are no-ops. If the first call does not actually quit and swap the binary, the flag stays set for the process lifetime and **no further install can succeed** — the app is pinned to its current version until restarted, and on restart it may come up on the old binary again.

## Why it matters beyond the annoyance

An app stuck on an old version keeps managing a home written by a newer one. See the companion issue on the missing downgrade guard — that combination is what makes this user-visible as "it rolled back for no reason."

## What to investigate

- Why `quitAndInstall` is being called more than once. A retry path, a user-triggered install on top of an auto-install, or a failed quit that leaves the flag set.
- For AppImage specifically: whether the update writes a new file beside the old one and leaves the launcher/desktop entry pointing at the old path. This machine has ~15 `OpenPalm-*.AppImage` files in `~/Downloads`, including `OpenPalm-0.13.0.AppImage` (2026-08-31 11:31) alongside older ones, and the running process is mounted from an AppImage that reports 0.12.42.
- Whether the same double-call happens on macOS, where the swap mechanism differs — the reporter sees the same behavior there, which argues the bug is in the call sequence rather than in AppImage handling.

## Acceptance

- `quitAndInstall` is called at most once per downloaded update, and a failure to quit is logged loudly rather than silently disabling all future installs.
- After an update, a normal relaunch starts the NEW version. A log assertion or a small integration check would keep this honest.

### LINKED ISSUE #636: No downgrade guard: an older app manages a newer OP_HOME and would pin the whole stack to its own older version
Found while diagnosing a stack stuck on rollback pins. This is the mechanism most likely behind "it rolls back for no reason" on both Linux and macOS.

## The state that exists right now on a real install

- `~/.openpalm/.skeleton-version` = **0.13.0**, `state/schema-version` = **11**
- The running desktop app = **v0.12.42** (see the companion self-updater issue)

So a 0.12.42 control plane is managing a home written by 0.13.0. I could not find any guard that refuses, warns, or degrades in that situation.

## Why it is dangerous, concretely

`advanceManagedImageVersions` (packages/lib/src/control-plane/versions.ts) advances pins to `targetPlatformVersion`, which defaults to the **PLATFORM_VERSION of the running code**:

```ts
if (value.startsWith('rollback-')) {
  const target = key === 'OP_VOICE_VERSION' ? VERSION_DEFAULTS.OP_VOICE_VERSION : targetPlatformVersion;
  updates[key] = target;
  updates[markerKey] = target;
}
```

If the operator clicks Update in the 0.12.42 app and the deploy SUCCEEDS, that writes `OP_ASSISTANT_VERSION=0.12.42` (and guardian/portal likewise) into `state/stack.env` — silently downgrading the entire stack to images from an older release. `applyHomeSeed`/`overwriteSystemTree` would additionally overwrite `system/` with the 0.12.42 skeleton, reverting managed compose files under a home whose schema stamp says 11.

Nothing about that presents to the user as a downgrade. It presents as "my stack went backwards on its own."

## What to build

A version-skew guard on the lifecycle entry points (`performUpgrade`, and whatever the desktop app calls on launch):

- Compare the running `PLATFORM_VERSION` against `.skeleton-version` (and/or `state/schema-version` against `HOME_SCHEMA_VERSION`).
- When the home is NEWER than the running code: refuse to apply managed files or advance version pins, and surface an actionable message ("this home was written by 0.13.0; you are running 0.12.42 — update the app first"). Read-only operation should still work.
- Decide explicitly whether a newer app on an older home (the normal upgrade direction) stays unguarded — it should.

## Acceptance

- An older app cannot rewrite `system/`, cannot advance version pins, and cannot stamp `.skeleton-version` backwards on a home newer than itself.
- A test covering the skew case, since this one is invisible until it has already happened.

### LINKED ISSUE #639: Failed updates never release the rollback- version pin; no CLI un-pin route, no UI visibility
## Summary

A failed `openpalm update` (any path — CLI or UI) leaves the stack permanently pinned to `rollback-<generation>` image tags in `state/stack.env`. The only code path that clears a `rollback-` pin is a **subsequent successful** `performUpgrade`. If every retry keeps failing, the pin never clears and every failed retry replaces it with a *newer* rollback generation, so the operator accumulates orphaned `rollback-*` images forever while staying stuck on old code. The pinned state is not surfaced anywhere in the UI or CLI.

## Mechanism (verified against `release/0.13.1`)

- `restoreRunningImageIds` (`packages/lib/src/control-plane/image-snapshots.ts:34-61`) re-tags the previously-running image IDs as `<namespace>/<image>:rollback-generation-<ts>-<pid>-<n>` and calls `patchStateEnvFile` to point `OP_ASSISTANT_VERSION`/`OP_GUARDIAN_VERSION`/`OP_PORTAL_VERSION`/`OP_VOICE_VERSION` at those tags (`image-snapshots.ts:58,60`).
- It is invoked as the `preserveImages` callback inside `runWithSnapshotRollback`'s catch handler (`packages/lib/src/control-plane/lifecycle.ts:344-348`), i.e. on **every** failed `performUpgrade` attempt (`lifecycle.ts:438-448`, `imageSnapshot` captured at line 448, `preserveImages` wired at the `performUpgrade` call site further down).
- The only arm that clears a `rollback-` value is `advanceManagedImageVersions` (`packages/lib/src/control-plane/versions.ts:133-136`):
  ```ts
  if (value.startsWith('rollback-')) {
    const target = key === 'OP_VOICE_VERSION' ? VERSION_DEFAULTS.OP_VOICE_VERSION : targetPlatformVersion;
    updates[key] = target;
    updates[markerKey] = target;
  }
  ```
- `advanceManagedImageVersions` is called from exactly one place in production code: `applyManagedFiles` (`lifecycle.ts:298`), which itself is invoked by `applyInstall`, `applyUpdate`, and `performUpgrade`. **Correction to the working theory**: it is *not* reached only via `performUpgrade` — `applyManagedFiles` runs inside the same `runWithSnapshotRollback` `run()` callback as the rest of the upgrade (`lifecycle.ts:449-455`), *before* `activateStack` is attempted. So on a failed attempt, the sequence within one `performUpgrade` call is: unpin (advanceManagedImageVersions clears the stale rollback- value left by the *previous* failure) → attempt `activateStack` → fail → `restoreRunningImageIds` re-pins to a brand-new `rollback-<generation>`. Net effect across repeated failures is exactly the reported bug: the pin never actually clears because the unpin and re-pin happen inside the same failing attempt, but the mechanism is "unpin-then-immediately-re-pin-on-failure", not "unpin never runs."
- `applyUpdate` (`lifecycle.ts:402-421`) is dead code in production — grep confirms it is only referenced from `packages/lib/src/index.ts` (export) and test files. Both the CLI (`packages/cli/src/commands/update.ts:40`) and the UI route (`packages/ui/src/routes/api/host/update/+server.ts:84`) call `performUpgrade` directly, so `applyUpdate`'s independent existence doesn't change the analysis.
- `openpalm rollback` (`packages/cli/src/commands/rollback.ts`) restores the pre-update **config snapshot** and re-runs `compose up`, but never touches `advanceManagedImageVersions` or the version markers — it cannot clear a rollback pin either.

## What establishes/refutes the reported description

1. **Confirmed**: a successful `performUpgrade` is the only path that clears a `rollback-` pin. No CLI command and no other API route calls `advanceManagedImageVersions`.
2. **No supported escape route in the CLI.** There is no `openpalm rollback --clear`, `openpalm unpin`, or equivalent. `writeVersions` (the operator-pin API, `versions.ts:168-170`, blanks `OP_MANAGED_*` markers) is called from exactly one production call site: `PATCH /api/host/versions` (`packages/ui/src/routes/api/host/versions/+server.ts:95`). Nothing in `packages/cli/src` calls `writeVersions`.
3. **There IS a supported UI escape route, but it's undiscoverable.** The Updates tab's "Advanced image tags" `<details>` panel (`packages/ui/src/lib/components/admin/updates/UpdatesTab.svelte:315-357`) PATCHes `/api/host/versions` → `writeVersions`, which accepts any string unconditionally and blanks the marker — it does not care that the current value starts with `rollback-`. So a UI operator *can* type `latest` (or an explicit tag) into all four fields and clear the pin without touching any file. But: the panel is collapsed by default, gives no indication the current values are rollback tags rather than normal ones, and the placeholder copy ("Most installations should keep these values set to `latest`") gives no hint that anything is wrong. A CLI-only/headless operator has no equivalent at all — for them the only route is hand-editing `state/stack.env`.
4. **Silent pin, confirmed.** Nothing surfaces the pinned state. `GET /api/host/versions` (`versions/+server.ts:29-36`) returns the raw `readVersions()` values with no flag distinguishing a `rollback-*` tag from an intentional pin. `UpdatesTab.svelte`'s "Compose containers" table shows the *running* image (which correctly reads `rollback-generation-...`, e.g. as `<code>`), but nothing calls it out. `RecoveryTab.svelte` (the "Recovery" tab: backups, secret-strip notice, stuck-lock notice) has no rollback-pin awareness at all despite being the natural home for it. `openpalm status` does not read version state either.

## Observed evidence (this machine, `~/.openpalm`, compose project `splinter`)

`state/stack.env` currently has all four services pinned:
```
OP_ASSISTANT_VERSION=rollback-generation-1788212586188-217761-1
OP_VOICE_VERSION=rollback-generation-1788212586188-217761-1
OP_GUARDIAN_VERSION=rollback-generation-1788212586188-217761-1
OP_PORTAL_VERSION=rollback-generation-1788212586188-217761-1
OP_MANAGED_ASSISTANT_VERSION=
OP_MANAGED_GUARDIAN_VERSION=
OP_MANAGED_PORTAL_VERSION=
OP_MANAGED_VOICE_VERSION=
```
`~/.config/@openpalm/electron/logs/main.log` shows three consecutive failed `performUpgrade` attempts, each producing a new rollback generation:
```
2026-08-31T17:14:55Z  update failed  "container for service guardian not found after up; container for service discord not found after up"
2026-08-31T17:22:08Z  update failed  "container for service discord not found after up"
2026-08-31T21:48:48Z  update failed  "container for service discord not found after up"
```
`docker images` shows the matching orphaned tags (`assistant`/`guardian`/`portal`/`voice`, one generation per failure), none ever removed:
```
openpalm/assistant:rollback-generation-1788196294580-1229079-1
openpalm/assistant:rollback-generation-1788196601273-1229079-2
openpalm/assistant:rollback-generation-1788212586188-217761-1   <- currently live pin
```
This is not a one-off: `docker images` on this host shows the same accumulation pattern going back to June 2026 (9 prior `rollback-generation-*` tags per image across 5 earlier failure clusters, e.g. `-1785309886682-11144-1`, `-1786552699130-2026317-3`), none pruned. The stack is currently up and healthy (`docker ps`: all `splinter-*` containers `Up ... (healthy)`) — the automatic rollback itself works correctly; the problem is purely that the pin never releases and nothing tells the operator.

Related but distinct: issue #636 ("No downgrade guard...") found the same `advanceManagedImageVersions` function unsafe in the *other* direction (an older app clearing a pin and advancing it backwards). Issue #581 ("Reduce Container Storage Growth", closed) flagged rollback/backup storage growth in general but didn't identify this specific unpruned-`rollback-*`-image-tag accumulation.

## Candidate fixes considered (not implemented)

1. **Clear the pin on a successful `openpalm start`/reconcile, not only on update.** Would help operators who give up retrying updates and just want the stack running again, but "successful start" of a stack already running on `rollback-*` tags doesn't mean those tags are safe to advance to `targetPlatformVersion` — the whole reason they're pinned is that the *newer* version failed. Auto-advancing on a plain start would silently re-attempt the same failing upgrade path outside of an explicit update action. Not safe as-is.
2. **`openpalm rollback --clear` / `openpalm unpin` command.** Directly addresses the CLI gap (finding 2). Needs to decide what it advances *to* — same `targetPlatformVersion` logic as `advanceManagedImageVersions`, or the previous pre-upgrade value. Cleanest, most explicit fix; low risk since it's operator-invoked.
3. **Surface the pinned state in the UI with a one-click clear.** Directly addresses the silent-pin gap (finding 4) and the discoverability gap (finding 3) — e.g. a `RecoveryTab.svelte` banner when any `readVersions()` value starts with `rollback-`, with a button that calls the existing `PATCH /api/host/versions` with the target tags. This doesn't require new lib code, just UI wiring, and reuses the already-supported `writeVersions` path.
4. **Have the rollback record the version it rolled back FROM.** Would remove the ambiguity in what "clear" should mean (recheck: `restoreRunningImageIds` currently only records the image ID/tag being demoted, not what it was replacing). Must not interact with the `writeVersions`/`writeManagedVersions` marker contract: `writeVersions` is the OPERATOR-PIN API and intentionally blanks `OP_MANAGED_*` to mean "never auto-advance this." A rollback-recorded "rolled back from X" value must not be written through `writeVersions` (that would make the rollback tag look like a deliberate operator pin and permanently exempt it from `advanceManagedImageVersions`) — it would need its own marker or reuse `writeManagedVersions`, which stamps the marker to match so a later `advanceManagedImageVersions` still recognizes and advances it. Any fix that touches this must preserve that an operator who deliberately pins via the Advanced image tags form is never auto-unpinned by (3)'s clear button or the accumulation cleanup below.

Independently, the orphaned `rollback-*` image tags (never pruned — 12 generations per image observed on this host) should be addressed regardless of which pin-clearing fix ships, since they accumulate even when the pin story is fixed.

## Acceptance criteria

- An operator whose updates keep failing has a documented, supported way to un-pin the stack from `rollback-*` tags without hand-editing `state/stack.env` — from the CLI, not only the UI.
- The UI surfaces (does not require expanding a collapsed "Advanced" panel to discover) that the running stack is on a rollback-pinned image, distinct from a deliberate operator pin.
- Whatever clear mechanism ships does not clear a pin an operator set deliberately via the Advanced image tags form / `PATCH /api/host/versions` (i.e. respects the `OP_MANAGED_*` marker contract already documented on `writeVersions`/`writeManagedVersions`).
- A test reproduces "N consecutive failed `performUpgrade` calls leave the stack pinned to the Nth rollback generation with no supported non-UI clear path" and passes once fixed.

### LINKED ISSUE #641: openpalm update aborts at teardown: cannot rm .system-previous-* because guardian writes node_modules as root
Found while upgrading a 0.12.x instance to 0.13.0 with `openpalm update`. Suggesting for 0.13.1.

## Symptom

`openpalm update` completes its migrations, then aborts at the very end:

```
EACCES: permission denied, rm '<OP_HOME>/.system-previous-<ts>-<pid>'
If something went wrong, your previous state was backed up before this update — run `openpalm rollback` to restore it.
```

The migrations themselves had already been applied and persisted, so the run is neither fully applied nor cleanly rolled back.

## Cause

The `.system-previous-*` staging directory holds the outgoing system tree, which includes `guardian/node_modules`. Those files are written by the guardian container running as root, so they end up root-owned on the host bind mount. The CLI runs as the invoking (non-root) user and cannot `rm -rf` them.

In the instance I hit this on, 3,926 entries under `.system-previous-*/guardian/node_modules` were root-owned.

## `repair-ownership` does not cover it

The obvious remedy is a no-op, because the staging directory is not a tracked bind mount:

```
$ openpalm repair-ownership
{"service":"lib:ownership-reconcile","msg":"Ownership already reconciled for uid=1000 — skipping repair walk"}
Ownership repair complete.
```

## Workaround

`sudo rm -rf <OP_HOME>/.system-previous-*` and re-run `openpalm update`. It recurs on each subsequent update that has to stage a system tree containing root-owned container output.

## Suggested fixes

- Include `.system-previous-*` (and any other CLI-created staging paths) in the `repair-ownership` walk.
- Or have the staging teardown chown/`rm` via the same privileged path that created the files.
- Or avoid staging `node_modules` at all — it is regenerated, not state worth preserving.

### LINKED ISSUE #642: Any operator-created root-owned file in OP_HOME aborts openpalm update at backup staging (EACCES copyfile)
Found while upgrading a 0.12.x instance to 0.13.0. Suggesting for 0.13.1. Related to #641 (same root cause class, different code path).

## Symptom

`openpalm update` aborts while staging its backup snapshot:

```
EACCES: permission denied, copyfile
  '<OP_HOME>/state/stack.env.bak-portfix-<ts>'
  -> '<OP_HOME>/data/backups/.staging-<ts>/state/stack.env.bak-portfix-<ts>'
```

## Cause

The backup staging step copies everything under `OP_HOME`. Any file an operator created with `sudo` — a routine `cp x x.bak` before hand-editing `stack.env`, for example — is root-owned, and the CLI runs unprivileged. One such file anywhere in the tree aborts the entire update.

Operators are actively encouraged to hand-edit files under `OP_HOME` (`config/stack/custom.compose.yml`, `state/stack.env`, `knowledge/env/*`), and several of those are root-owned to begin with, so reaching for `sudo` is natural. The failure gives no indication that an unrelated backup file is the blocker.

## Workaround

Move every operator-created `.bak`/`.pre-*` file out of `OP_HOME` before updating.

## Suggested fixes

- Skip obvious operator backup artifacts (`*.bak`, `*.bak-*`, `*.pre-*`) when staging — they are not state the upgrade needs.
- Fail with a specific, actionable message naming the offending path and suggesting removal, rather than a raw `EACCES`.
- Or make the backup staging tolerant: warn and continue on individual unreadable files instead of aborting.

### LINKED ISSUE #643: Port migration reassigns a host port already bound by another OpenPalm instance, and reverts operator overrides each run
Found while upgrading a 0.12.x instance to 0.13.0 on a host running three OpenPalm instances side by side. Suggesting for 0.13.1.

## Symptom

`openpalm update` fails at the reapply step. After several rounds of diagnosis the underlying cause was:

```
Error response from daemon: failed to set up container networking:
driver failed programming external connectivity on endpoint <inst>-assistant-1:
Bind for 127.0.0.1:3810 failed: port is already allocated
```

The port was held by a **different OpenPalm instance's** assistant container on the same host, publishing its UI on that port.

## Cause

The port migration assigns default host ports per instance:

```
{"service":"config-persistence","msg":"Migrated default host port assignments","extra":{"assistant":"3810","ui":"3800"}}
```

It does not check whether the port is already bound by another instance on the host, and it **reassigns on every run** — a manual override in `state/stack.env` is silently reverted, so the operator cannot work around it locally:

```
# set OP_ASSISTANT_PORT=3811 by hand, run openpalm update
$ grep OP_ASSISTANT_PORT state/stack.env
OP_ASSISTANT_PORT=3810   # reverted
```

The only way out was to move the *other* instance's port, which meant taking a healthy production instance down to unblock the one being upgraded.

## Note

`openpalm doctor` does report it, but ambiguously — a live instance always shows its own ports as in use, so a genuine cross-instance collision is indistinguishable from normal operation:

```
Ports:
  ui (3800): available
  assistant (3810): CONFLICT
```

## Suggested fixes

- Check host-wide port availability during the migration and pick the next free port in the instance's own range.
- Treat an explicit operator value in `state/stack.env` as authoritative instead of reverting it.
- Have `doctor` distinguish "in use by this instance" from "in use by another process", so the warning is actionable.

### LINKED ISSUE #644: Reapply failures surface only the Compose progress line ("Container X Recreate"), swallowing the Docker daemon error
Found while upgrading a 0.12.x instance to 0.13.0. Suggesting for 0.13.1. This is what made #643 expensive to diagnose.

## Symptom

When the reapply step fails, the entire underlying Docker error is reduced to the container name and the word `Recreate`:

```
{"service":"lifecycle","msg":"failed to reapply restored stack","extra":{"error":"Container <inst>-assistant-1 Recreate"}}
Container <inst>-assistant-1 Recreate Additionally, automatic rollback did not fully recover: restored stack reapply failed: Container <inst>-assistant-1 Recreate
```

That string is the Compose *progress line*, not the error. There is nothing to act on: no reason, no exit status, no daemon message.

## What was actually wrong

Reproducing the same recreate by hand immediately gave the real cause:

```
$ docker compose --project-name <inst> -f ... up -d assistant
Error response from daemon: failed to set up container networking:
driver failed programming external connectivity on endpoint <inst>-assistant-1:
Bind for 127.0.0.1:3810 failed: port is already allocated
```

A port bind failure — diagnosable in seconds if surfaced, but it took several failed update cycles to find, each leaving the stack down.

## Impact

The stack is left with containers created-but-not-started and the service down, while the message offers no path forward other than `openpalm rollback`.

## Suggested fix

Propagate stderr from the failing `docker compose` invocation into the lifecycle error rather than the progress line. Even truncated daemon output would have made this self-service.

### LINKED ISSUE #645: 0.12→0.13 upgrade silently drops akm profiles.llm, leaving the instance with no LLM engine (validate still reports OK)
Found while upgrading a 0.12.x instance to 0.13.0. Suggesting for 0.13.1.

## Symptom

After a successful `openpalm update`, the instance's akm config is correctly migrated to `configVersion: 0.9.0` — but `engines` is empty, so the assistant has **no LLM engine configured at all**. Nothing in the update output indicates this.

Before (0.12.x, legacy shape):

```json
{ "profiles": { "llm": { "default": { "endpoint": "...", "model": "..." } } } }
```

After (0.13.0):

```json
{ "configVersion": "0.9.0", "bundles": {...}, "engines": {} }
```

## Cause

This is akm behaving as designed — 0.9.0 retires `profiles` and its own error text is explicit that the translation is not automatic:

> `profiles` is retired in 0.9; configure engines and improve.strategies instead

> Recreate engines and improve.strategies manually for AKM 0.9.0; profile-based configuration is not translated automatically.

The gap is that OpenPalm drives this migration during `update` and drops the key silently. akm would have failed closed and told the operator; because OpenPalm migrates the file first, the instance comes up "healthy" with a config that validates and an assistant that has no engine.

`openpalm validate` reports `Configuration OK.` in this state.

## Suggested fixes

- Translate `profiles.llm.<name>` → `engines.<name>` during the upgrade — the fields map almost one-to-one (`endpoint`, `model`, `provider`, `apiKey`).
- Or, if translation is deliberately out of scope, emit a loud warning naming what was dropped, and have `validate`/`doctor` flag an empty `engines` as a defect rather than OK.

itlackey:
## Summary

Ten defect fixes for the 0.13.1 milestone, one commit per issue (two for #643 and for #641/#642), no features or refactors. Each fix carries a regression test that was shown to fail without it, and each was reproduced before/after against a real isolated `OP_HOME` through the real CLI/lib code paths.

Closes #645, closes #643, closes #641, closes #642, closes #639, closes #636, closes #635, closes #644, closes #632, closes #633.

| Issue | Fix |
|---|---|
| #645 | `stripRetiredAkmKeys` now translates `profiles.llm.*` → akm 0.9 `engines.*` (endpoint via the idempotent `buildAkmEndpoint`, model, provider, env-var `apiKey` refs) before the strip; untranslatable fields are named in a warning; `openpalm validate` warns on zero engines. |
| #643 | `migrateLegacyDefaultPorts` carries an explicit `OP_ASSISTANT_PORT`/`OP_UI_PORT` from the consolidated `state/stack.env` through the chain (a rollback's restored schema-version re-runs it and consolidation let the legacy default win), and probes host-wide availability before writing a fresh default; `migrateConsolidatedDefaultPorts` skips the 3800→3810 swap when 3810 is held by someone else. Migrations became async; the ripple is mechanical (`await` at four call sites, top-level `await` in the UI server hooks). |
| #641 | Still reproduced on a first 0.12→0.13.1 upgrade (old root-owned guardian `node_modules` gets staged). `overwriteSystemTree`'s final `rmSync` of `.system-previous-*` is now best-effort: warn naming the path instead of failing a completed update. |
| #642 | `backupOpenPalmHome` walks the tree itself: an unreadable file outside the four paths `openpalm rollback` restores is skipped and recorded in `.backup-complete`; one of those four fails loud with the path and a `chown` hint. |
| #639 | `clearRollbackPins()` in lib (clears only `rollback-` prefixed values, re-stamps `OP_MANAGED_*`, never touches an operator pin); `openpalm unpin`; `POST /api/host/versions/clear-rollback-pin`; Recovery-tab banner with one-click clear. |
| #636 | `assertHomeNotNewerThanApp()` compares `.skeleton-version`/`state/schema-version` against the running build and refuses `applyInstall`/`applyUpdate`/`performUpgrade`/`applyHomeAssets`/`clearRollbackPins` before any snapshot or write; `status`/`validate`/`start` unaffected; missing or non-semver stamps never trip it. |
| #635 | `DesktopUpdater` gets a per-download `installAttempted` guard so electron-updater's `install()` is never called twice (the before-quit handler was re-entering it after `quitAndInstall()`); a failed quit-time install is logged loudly. |
| #644 | `COMPOSE_PROGRESS_RE` was missing bare `Recreate`; `applyStack`'s post-failure branch discarded compose stderr once `ps -a` found a row. Both fixed in the shared derivation; verified against a real daemon port-bind error. |
| #632 | `auditResolvedComposeSecrets` checks each resolved secret file exists (compose prunes secrets to active services, so no addon false-positives) and names the secret, path, and remedy. |
| #633 | `bun install --lockfile-only` never rewrites a workspace's own version field (the 0.13.0 CI prep commit proves it). `scripts/sync-lockfile-versions.mjs` patches exactly those fields; wired into the release stamp step; lockfile trued up. |

## Verification

- `bun run lint` clean; `bun run test` (CI's command, `--isolate`): 2609 pass / 0 fail; `bun run ui:check`: 0 errors; electron: 188 pass; `bun run --cwd packages/electron typecheck` clean.
- Tests requiring an unprivileged user (#641/#642) were run as a non-root user here; they self-skip under uid 0.
- Environment limits, stated honestly: this sandbox has a Docker daemon and Compose v5 but Docker Hub image pulls are policy-blocked, so no fix was verified on a booted stack. Every fix was instead reproduced and confirmed on a real isolated `OP_HOME` through the real CLI/lib entry points (`runHomeMigrations`, `stripRetiredAkmConfigKeys`, `backupOpenPalmHome`, `overwriteSystemTree`, `applyStack` against the real daemon, `openpalm unpin`/`update`/`start`/`validate`). `./scripts/akm-pin-integration-smoke.sh` could not run for the same reason; in its place the pinned akm-cli 0.9.6 and 0.9.7 binaries were installed directly and `akm config list`/`akm health` were run against the #645 fixture before and after. A live-stack pass on the operator's side is still warranted before tagging.
- Note for maintainers: with bun 1.3.11, `bun test --isolate` on this suite shows 161 cross-file mock-leak failures even on the untouched base; under CI's pinned bun 1.3.14 it is fully green.

https://claude.ai/code/session_018rQ2BpDPUCde6Na4gZdKpu

---
_Generated by [Claude Code](https://claude.ai/code/session_018rQ2BpDPUCde6Na4gZdKpu)_

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,6 +111,7 @@ jobs:
           node scripts/bump-unit.mjs
           UNIT=electron node scripts/bump-unit.mjs
           bun install --lockfile-only
+          node scripts/sync-lockfile-versions.mjs
       - name: Commit and bundle candidate
         id: bundle
         env:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,51 @@ Versioning follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- **`openpalm update` from 0.12.x no longer silently drops the akm LLM
+  engine (#645).** The retired `profiles.llm.*` block is translated into akm
+  0.9 `engines.*` (endpoint, model, provider, and an env-var `apiKey`
+  reference) before the retired keys are stripped; anything untranslatable is
+  named in a warning, and `openpalm validate` now warns when an akm config has
+  zero engines instead of reporting OK.
+- **Port migrations no longer revert an operator's `state/stack.env` port or
+  assign a port another instance already holds (#643).** An explicit
+  `OP_ASSISTANT_PORT`/`OP_UI_PORT` in the consolidated file survives a re-run
+  of the migration chain (which a rollback's restored schema-version
+  triggers), and a fresh default is probed host-wide and steps to the next
+  free port when a sibling install binds it.
+- **Root-owned files in `OP_HOME` no longer abort `openpalm update` (#641,
+  #642).** The backup snapshot skips an unreadable file it does not need and
+  records it in `.backup-complete`, failing loud only for the files
+  `openpalm rollback` restores; a leftover root-owned `.system-previous-*`
+  tree (a pre-0.13.1 guardian's `node_modules`) is warned about instead of
+  failing the already-completed update.
+- **Failed updates can be un-pinned from `rollback-generation-*` image tags
+  (#639).** New `openpalm unpin` command and a Recovery-tab banner with a
+  one-click clear share one lib function that clears only `rollback-`
+  prefixed values, so a deliberate operator pin is never touched.
+- **An older app refuses to manage a newer `OP_HOME` (#636).** Install,
+  update, upgrade, home-asset seeding, and `unpin` compare
+  `.skeleton-version` and `state/schema-version` against the running build
+  and refuse with an actionable message instead of rewriting `system/` or
+  advancing image pins backwards; `status`, `validate`, and `start` keep
+  working.
+- **The desktop self-updater never calls electron-updater's `install()`
+  twice (#635).** A second call after a quit-time install was a silent no-op
+  that jammed every later install for the process; it is now refused, and a
+  failed quit-time install is logged loudly.
+- **Reapply and deploy failures surface the Docker daemon's real error
+  (#644).** The bare `Recreate` Compose progress line is filtered out, and a
+  per-service failure reason now carries the summarized compose stderr
+  instead of a templated "did not become healthy".
+- **A home missing a compose secret file fails before Docker does (#632).**
+  The pre-activation secret audit checks that each resolved secret file
+  exists and names the secret, the path, and the remedy, instead of Docker's
+  opaque bind-mount error after `install --no-start`.
+- **`bun.lock` workspace versions are trued up at release stamp time
+  (#633).** `bun install --lockfile-only` never rewrites a workspace's own
+  version field; the release workflow now patches exactly those fields after
+  the bump.
+
 - **Preferred-model saves from the admin UI now persist — and clearing one
   actually clears it.** Three defects in the `opencode.json` write path
   (`packages/ui/src/lib/server/opencode/config.ts`), found together. First:
--- a/bun.lock
+++ b/bun.lock
@@ -10,7 +10,7 @@
     },
     "packages/cli": {
       "name": "openpalm",
-      "version": "0.13.0-beta.31",
+      "version": "0.13.0",
       "bin": {
         "openpalm": "./bin/openpalm.js",
       },
@@ -25,7 +25,7 @@
     },
     "packages/electron": {
       "name": "@openpalm/electron",
-      "version": "0.13.0-beta.31",
+      "version": "0.13.0",
       "dependencies": {
         "electron-updater": "^6.8.9",
       },
@@ -53,7 +53,7 @@
     },
     "packages/lib": {
       "name": "@openpalm/lib",
-      "version": "0.13.0-beta.31",
+      "version": "0.13.0",
       "dependencies": {
         "dotenv": "^17.4.2",
         "multicast-dns": "^7.2.5",
@@ -99,11 +99,11 @@
     },
     "packages/skeleton": {
       "name": "@openpalm/skeleton",
-      "version": "0.13.0-beta.31",
+      "version": "0.13.0",
     },
     "packages/ui": {
       "name": "@openpalm/ui",
-      "version": "0.13.0-beta.31",
+      "version": "0.13.0",
       "devDependencies": {
         "@axe-core/playwright": "^4.12.1",
         "@eslint/compat": "^2.1.0",
--- a/docs/managing-openpalm.md
+++ b/docs/managing-openpalm.md
@@ -312,6 +312,27 @@ offline; `{"upgraded":false}` means there was nothing to do. Boot never runs
 this for you: applying a destructive migration stays an explicit operator
 action, the same rule that keeps your task files unrewritten at boot.
 
+### A failed update leaves the stack pinned to a rollback image
+
+**A failed update pins your images to a preserved rollback tag, not to
+`latest`.** Before a failed update can mutate anything, OpenPalm retags each
+running image as `<namespace>/<service>:rollback-generation-<id>` and points
+`state/stack.env` at those tags, so the automatic recovery above restarts the
+exact images that were running before — never a `latest` that might not match
+your restored config. If `update` keeps failing, every attempt repeats this,
+so the stack stays correctly pinned to whichever rollback generation is
+newest; it never drifts to a stale one, but the pin also never clears itself.
+Run `openpalm unpin` once the underlying problem is fixed (bad connectivity, a
+bad tag, disk space, …) to release the pin and let the next `update`/`start`
+pull the normal release tag again — it only edits `state/stack.env` and never
+touches a version you set yourself (a value without the `rollback-` prefix),
+so run `update` afterward to actually apply it:
+
+```bash
+openpalm unpin
+openpalm update
+```
+
 ### Desktop app updates
 
 The desktop app updates as one complete application — shell and UI together —
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -272,6 +272,24 @@ openpalm rollback
 Raw copying is incomplete because install/update also generates state, private
 secrets, caches, and runtime files.
 
+### Stuck on a rollback image tag after repeated failed updates
+
+Each failed `update` retags your previous images as
+`rollback-generation-<id>` and pins `state/stack.env` to them so the
+automatic recovery restarts exactly what was running before — but nothing
+clears that pin on its own, so repeated failures just move it to a newer
+generation. Once the update failure itself is fixed, release the pin and
+apply the real release tag:
+
+```bash
+openpalm unpin
+openpalm update
+```
+
+`openpalm unpin` only clears `state/stack.env` keys that are still on a
+`rollback-generation-*` tag; it never touches an image tag you pinned
+yourself.
+
 ## Factory Reset
 
 Back up first. Then remove the stack and all OpenPalm-owned trees:
--- a/packages/cli/src/commands/install.ts
+++ b/packages/cli/src/commands/install.ts
@@ -413,7 +413,7 @@ async function prepareInstallFiles(
 	// runs (deploy happens later from inside the UI), so this explicit seed is
 	// the only one that runs before the wizard comes up.
 	await seedSkeletonFromEmbedded(applyHomeSeed, homeDir, dataDir);
-	runHomeMigrations(homeDir);
+	await runHomeMigrations(homeDir);
 	// Materialize the embedded UI build into data/ui — no network, no backup:
 	// this binary's build wins unconditionally once its stamp differs from
 	// what's already there (a no-op on a repeat install at the same version).
--- a/packages/cli/src/commands/unpin.test.ts
+++ b/packages/cli/src/commands/unpin.test.ts
@@ -1,0 +1,147 @@
+/**
+ * #639 — `openpalm unpin` clears rollback-generation-* pins left behind by a
+ * failed update via the shared clearRollbackPins() lib function, and never
+ * touches a genuine operator pin.
+ *
+ * Harness: real @openpalm/lib + a seeded temp OP_HOME, mirroring
+ * reset-password.test.ts's documented convention (mocking the shared
+ * @openpalm/lib import can leak across files sharing this bun test process,
+ * so the package/cli-state mocks are re-asserted at the START of every test).
+ */
+import { afterEach, beforeEach, describe, expect, mock, test } from 'bun:test';
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import * as realLib from '../../../lib/src/index.ts';
+import * as realCliState from '../lib/cli-state.ts';
+
+const cliStateModuleUrl = new URL('../lib/cli-state.ts', import.meta.url).href;
+const unpinModuleUrl = new URL('./unpin.ts', import.meta.url).href;
+const mainModuleUrl = new URL('../main.ts', import.meta.url).href;
+
+const originalHome = process.env.OP_HOME;
+let tempHome: string;
+
+function resetMocks(): void {
+	mock.restore();
+	mock.module('@openpalm/lib', () => ({ ...realLib }));
+	mock.module(cliStateModuleUrl, () => ({
+		...realCliState,
+		ensureValidState: () => {
+			const state = realLib.createState();
+			if (realLib.classifyLocalInstall(state.stackDir, state.homeDir) === 'not_installed') {
+				throw new Error(
+					'OpenPalm is not installed in this OP_HOME yet. Run `openpalm install` first.'
+				);
+			}
+			state.artifacts = realLib.resolveRuntimeFiles();
+			return state;
+		}
+	}));
+}
+
+function stackEnvPath(): string {
+	return join(tempHome, 'state', 'stack.env');
+}
+
+beforeEach(() => {
+	tempHome = mkdtempSync(join(tmpdir(), 'openpalm-unpin-'));
+	process.env.OP_HOME = tempHome;
+
+	// Seed a minimal "installed" OP_HOME so ensureValidState() doesn't refuse.
+	mkdirSync(join(tempHome, 'system', 'stack'), { recursive: true });
+	writeFileSync(join(tempHome, 'system', 'stack', 'core.compose.yml'), 'services: {}\n');
+	mkdirSync(join(tempHome, 'state'), { recursive: true });
+});
+
+afterEach(() => {
+	mock.restore();
+	mock.module('@openpalm/lib', () => ({ ...realLib }));
+	mock.module(cliStateModuleUrl, () => ({ ...realCliState }));
+	if (originalHome === undefined) delete process.env.OP_HOME;
+	else process.env.OP_HOME = originalHome;
+	rmSync(tempHome, { recursive: true, force: true });
+});
+
+describe('runUnpinAction (#639)', () => {
+	test('clears the operator-reported stack.env shape: rollback- values with blank OP_MANAGED_* markers', async () => {
+		writeFileSync(
+			stackEnvPath(),
+			[
+				'OP_ASSISTANT_VERSION=rollback-generation-1788212586188-217761-1',
+				'OP_VOICE_VERSION=rollback-generation-1788212586188-217761-1',
+				'OP_GUARDIAN_VERSION=rollback-generation-1788212586188-217761-1',
+				'OP_PORTAL_VERSION=rollback-generation-1788212586188-217761-1',
+				'OP_MANAGED_ASSISTANT_VERSION=',
+				'OP_MANAGED_GUARDIAN_VERSION=',
+				'OP_MANAGED_PORTAL_VERSION=',
+				'OP_MANAGED_VOICE_VERSION=',
+				''
+			].join('\n')
+		);
+
+		resetMocks();
+		const { runUnpinAction } = await import(`${unpinModuleUrl}?t=${Math.random()}`);
+		await runUnpinAction();
+
+		const versions = realLib.readVersions(realLib.createState());
+		expect(versions.OP_ASSISTANT_VERSION).toBe(realLib.PLATFORM_VERSION);
+		expect(versions.OP_GUARDIAN_VERSION).toBe(realLib.PLATFORM_VERSION);
+		expect(versions.OP_PORTAL_VERSION).toBe(realLib.PLATFORM_VERSION);
+		expect(versions.OP_VOICE_VERSION).toBe('latest');
+		for (const key of realLib.SERVICE_VERSION_KEYS) {
+			expect(versions[key]).not.toMatch(/^rollback-/);
+		}
+	});
+
+	test('never clears a genuine operator pin (no rollback- prefix), even with a blank managed marker', async () => {
+		writeFileSync(
+			stackEnvPath(),
+			[
+				'OP_ASSISTANT_VERSION=rollback-generation-1',
+				'OP_GUARDIAN_VERSION=my-operator-pinned-build',
+				'OP_MANAGED_ASSISTANT_VERSION=',
+				'OP_MANAGED_GUARDIAN_VERSION=',
+				''
+			].join('\n')
+		);
+
+		resetMocks();
+		const { runUnpinAction } = await import(`${unpinModuleUrl}?t=${Math.random()}`);
+		await runUnpinAction();
+
+		const content = readFileSync(stackEnvPath(), 'utf-8');
+		expect(content).toContain('OP_GUARDIAN_VERSION=my-operator-pinned-build');
+		expect(content).not.toContain('OP_ASSISTANT_VERSION=rollback-generation-1\n');
+	});
+
+	test('is a no-op (and says so) when nothing is pinned to a rollback generation', async () => {
+		writeFileSync(stackEnvPath(), 'OP_ASSISTANT_VERSION=0.13.0\n');
+
+		resetMocks();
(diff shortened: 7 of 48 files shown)


itlackey:
**CI note: the Quality gates failure on 7c33124 is not this PR's, and a fix is now ported in.**

- Failing step: `./scripts/guardian-image-offline-smoke.sh` → `FAIL: entrypoint did not use the baked guardian package`.
- Same failure on the base: release/0.13.1's own CI run for a000687 (run 33585261956) fails on the identical line. That commit deleted the guardian entrypoint's boot-time package override, and with it the `using image-baked Guardian package` echo the smoke greps for. The guardian itself boots the baked package correctly under `--network none` in both runs (the boot receipt shows `"entry":"/opt/openpalm/guardian-pkg/node_modules/@openpalm/guardian/src/server.ts"`).
- Ported fix (570bddc): the smoke now asserts on that structured boot receipt's `entry` path, which only the image layer can produce (the shipped compose bind-mounts `/opt/openpalm/guardian`, never `/opt/openpalm/guardian-pkg`). Verified the grep against the exact receipt line from the CI log and against a wrong path. The smoke cannot run in my sandbox (image pulls blocked), so CI on this push is the real check.

---

PR #660, opened 2026-09-02
fix: 0.13.1 design fixes and test lanes for the whole milestone (#641–#659)

### LINKED ISSUE #632: install --no-start leaves a home that cannot docker compose up

### LINKED ISSUE #633: bun.lock drifts from platform package.json versions across release cycles

### LINKED ISSUE #639: Failed updates never release the rollback- version pin; no CLI un-pin route, no UI visibility

### LINKED ISSUE #641: openpalm update aborts at teardown: cannot rm .system-previous-* because guardian writes node_modules as root

### LINKED ISSUE #642: Any operator-created root-owned file in OP_HOME aborts openpalm update at backup staging (EACCES copyfile)

### LINKED ISSUE #643: Port migration reassigns a host port already bound by another OpenPalm instance, and reverts operator overrides each run

### LINKED ISSUE #644: Reapply failures surface only the Compose progress line ("Container X Recreate"), swallowing the Docker daemon error

### LINKED ISSUE #645: 0.12→0.13 upgrade silently drops akm profiles.llm, leaving the instance with no LLM engine (validate still reports OK)

### LINKED ISSUE #648: Each failed upgrade retry writes a full undeduplicated home snapshot including workspace/.git and node_modules (4.2 GB in 8 minutes)
Found upgrading 0.12.x → 0.13.0. Suggesting for 0.13.1. Adjacent to the closed #499 (free-space pre-check) and #581 (container storage growth), but neither covers this: the problem is not *that* a backup is taken, it is that a **full, undeduplicated copy is taken per failed attempt**, and that it includes regenerable trees.

## What happened

An upgrade failed and was retried several times. Each attempt wrote its own snapshot. In an **8-minute window** (21:50:10 → 21:58:32) this produced **10 directories totalling 4.2 GB**:

```
2026-09-01T21-50-10-766Z-pre-rollback   324K
2026-09-01T21-51-15-064Z-pre-rollback   324K
2026-09-01T21-53-10-171Z-pre-rollback   324K
2026-09-01T21-54-22-531Z-pre-rollback   324K
2026-09-01T21-55-45-623Z-pre-rollback   125M
2026-09-01T21-56-09-822Z                1.9G   <-- full home incl. workspace/
2026-09-01T21-56-13-437Z                125M
2026-09-01T21-56-15-441Z-pre-rollback   324K
2026-09-01T21-58-27-700Z                1.9G   <-- full home incl. workspace/, again
2026-09-01T21-58-32-995Z                125M
```

**96% of the bulk (4.05 GB) is in the two full snapshots**, and both contain a byte-identical copy of the same 687 MB `workspace/dc-op-manual/.git` directory. On a 55 GB root partition that is 7.6% of the disk consumed by two copies of a git checkout that was never modified.

## Two distinct causes

**1. `workspace/` is not excluded.** The only exclusion list is:

```js
UNBACKED_TOP_LEVEL = new Set(["data", "cache"]);
```

So `workspace/` is copied verbatim — `.git`, `node_modules`, everything — with no `.gitignore` awareness. These are regenerable or independently version-controlled; they are not stack state worth snapshotting on every upgrade attempt.

**2. No deduplication between snapshots.** Verified: the same relative file (`system/assistant/node_modules/uuid/README.md`, identical 16,630-byte content) has **different inodes** across the two 1.9 GB snapshots (`1201376` vs `1187614`). Each is a genuine independent copy — no hardlink, no reflink, no incremental logic. Nothing changed between the two attempts, yet the full tree was copied twice.

## Suggested fixes

Any one of these would have avoided the 4 GB:

- Exclude `workspace/**/.git` and `workspace/**/node_modules` from the full-backup path (or add `workspace` to `UNBACKED_TOP_LEVEL` if it is not considered stack state).
- Hardlink (`cp -al`) or reflink unchanged files between successive snapshots.
- Skip re-snapshotting on a **retry** when the previous attempt's snapshot already captured an identical pre-attempt state — the 5 retries here all had the same starting state.

## Secondary, unrelated to size

`openpalm backups --help` includes a disk-space warning that tells the user to run `openpalm backups list`. That subcommand does not exist — `backups` accepts only `prune`. One-line doc fix, or add the command.

### LINKED ISSUE #649: Secret-boundary audit directs operators to *_FILE, but env-consuming tools (akm) need knowledge/env/user.env — which is never mentioned
Found upgrading 0.12.x → 0.13.0. Suggesting for 0.13.1. The audit itself is good and caught a real mistake of mine — this is about the remediation advice being wrong for the most likely case.

## What the audit says

Putting a credential in a service `environment:` block (via `config/stack/custom.compose.yml`) correctly fails:

```
Refusing Compose stack activation: secret-boundary audit failed.
compose-secret-env-var: service assistant environment key LAB_API_KEY is
secret-like; expose only LAB_API_KEY_FILE. (services.assistant.environment.LAB_API_KEY)
```

And the same key in `state/stack.env` is stripped, then refused:

```
stack-env-secret-key: /…/state/stack.env must not contain secret-like key
LAB_API_KEY; store it as a narrowly granted file secret and expose
LAB_API_KEY_FILE instead.
```

## Why the advice does not work here

Both messages direct the operator to `LAB_API_KEY_FILE`. But the consumer in this case is **akm**, whose schema *requires* `apiKey` to be an environment reference and rejects anything else:

```
apiKey: z.string().regex(ENV_REFERENCE_PATTERN, "apiKey must be $VAR or ${VAR}")
```

akm has no `*_FILE` convention — it resolves `$LAB_API_KEY` from `process.env` at call time. So following the audit's advice literally produces a variable akm cannot consume, and there is no hint that another path exists.

## What actually works

`knowledge/env/user.env` — which `core.compose.yml`'s own comments describe as sourced by the assistant entrypoint at startup:

> User-managed env config lives in the akm `env:user` file at `knowledge/env/user.env` … The assistant entrypoint sources that file at startup (akm `env path env:user`).

Putting `LAB_API_KEY=…` there satisfies the audit (it is neither a compose env key nor `stack.env`) **and** reaches akm's `process.env`. It took reading the compose comments to find this; neither audit message mentions it.

## Suggested fix

Extend both audit messages to name `env:user` as the correct destination for credentials consumed as environment variables by in-container tooling, reserving the `*_FILE` guidance for consumers that actually read file secrets. Something like:

> …store it as a narrowly granted file secret and expose `<KEY>_FILE`, or — if the consumer reads it from the environment — put it in `knowledge/env/user.env`, which the assistant entrypoint sources at startup.

Documenting the distinction in the upgrade notes would also help, since this only surfaces when an operator adds a third-party credential.

### LINKED ISSUE #650: openpalm.sh defaults to project name "openpalm" and the wrong compose/env paths, creating a duplicate parallel stack
Found on a host running three OpenPalm instances at 0.13.0. Low severity but it cost real time during an incident.

## The problem

`OP_HOME/openpalm.sh` describes itself as wrapping "the same `docker compose` invocation the OpenPalm CLI and admin UI use". On a real 0.13.0 install it does not, in three ways:

| | `openpalm.sh` uses | Actual deployment uses |
|---|---|---|
| Project name | `${OP_PROJECT_NAME:-${COMPOSE_PROJECT_NAME:-openpalm}}` | the value in `state/stack.env`, e.g. `rhiza` |
| Compose files | `$STACK_DIR/{core,services,channels,custom}.compose.yml` under `config/stack/` | `system/stack/{core,services,portals}.compose.yml` + `config/stack/custom.compose.yml` |
| env-file | `knowledge/env/stack.env` | `state/stack.env` |

The project-name default is the damaging one. `OP_PROJECT_NAME` lives in `state/stack.env`, but `openpalm.sh` only sources `knowledge/env/stack.env` — so it never sees it and silently falls back to the literal project name `openpalm`.

## What that does

Running `./openpalm.sh compose up -d assistant` against a live instance whose real project is `rhiza` does not restart that instance. It tries to create a **second, parallel** container `openpalm-assistant-1` alongside the running one, and dies on a port conflict:

```
Container openpalm-assistant-1  Creating
Container openpalm-assistant-1  Created
Container openpalm-assistant-1  Starting
Error response from daemon: … Bind for 0.0.0.0:3810 failed: port is already allocated
```

It also leaves an orphan `Created` container and an `openpalm_assistant-artifacts` volume behind, which the operator then has to notice and clean up. The live instance is untouched — so no outage — but the failure mode is confusing, and on a host where the port *was* free it would have started a duplicate stack against the same bind mounts.

The correct invocation for this install turned out to be:

```bash
docker compose --project-name rhiza \
  -f system/stack/core.compose.yml \
  -f system/stack/services.compose.yml \
  -f system/stack/portals.compose.yml \
  -f config/stack/custom.compose.yml \
  --env-file state/stack.env up -d assistant
```

## Suggested fix

Either update `openpalm.sh` to source `state/stack.env` (picking up `OP_PROJECT_NAME`) and to assemble the file list from `system/stack/` the way the control plane does, or — since it is explicitly an example — have it refuse to run when `OP_PROJECT_NAME` is unset rather than silently defaulting to a project name that may not exist. A one-line guard would prevent the duplicate-stack scenario entirely.

Context: the canonical `openpalm` CLI was not on `PATH` on this host, which is why the example helper got reached for in the first place. `install` placing a symlink, or the helper printing "use the openpalm CLI, found at X", would also help.

### LINKED ISSUE #651: EPIC: the whole test suite runs in one execution shape, and 7 of 8 recent escapes live outside it
Eight defects (#641 #642 #643 #644 #645 #648 #649 #650) were found in a single afternoon upgrading three 0.12.x instances to 0.13.0 on one host. None were caught by 2,566 test cases across 192 test files. This issue explains why, and tracks the two lanes that would close the gap.

## The suite is not under-invested. It runs in one execution shape.

Measured on the current tree:

| | |
|---|---|
| `.test.ts` files | 192 (2,566 `it()`/`test()` cases) |
| Files using `mkdtemp`/`mkdtempSync` | **123 of 192** (348 occurrences) |
| Files touching Docker / a real container | 16 |
| Shell scripts booting a real container | 6 |
| Tests exercising **two instances on one host** | **0** |
| Tests exercising **root-written files + unprivileged CLI** | **3, all of which no-op in CI** (see #653) |

Every one of those 2,566 cases runs as **one process, one uid, one fresh temp directory, one instance, no prior-release artifacts**. Seven of the eight defects live in seams that shape cannot express:

- #641, #642 — a file written by a container as root, deleted later by an unprivileged CLI
- #643, #650 — a second instance already holding a port / a project name on the same host
- #648 — accumulation across *repeated* failed attempts
- #645 — a config file two systems disagree about the shape of
- #644 — the actual byte stream `docker compose` emits during a **recreate** (vs. a create)

Adding more same-shape unit tests will not catch any of these. The unit layer is doing its job within its model; the model is the problem.

## Three artifacts that make this concrete

**1. A green test encodes the data loss as the specification.**
`packages/ui/src/routes/api/assistant/akm/server.vitest.ts:425` — *"strips every retired 0.8 key from a pre-upgrade config on the first PATCH"* seeds `profiles: { llm: { default: { endpoint: 'http://x' } } }`, then asserts `expect(config, key).not.toHaveProperty(key)` for `profiles`. It never checks the endpoint was translated into an `engines` entry. That is #645, written down as intended behaviour, passing on every run.

**2. Three tests can never fail.**
`packages/lib/src/control-plane/config-persistence-operator-ids.test.ts` guards its mixed-uid cases with `if (!isRootProcess()) return;` (lines 53, 70, 95). `ci.yml` runs `bun run test` on plain `ubuntu-latest` as a normal user. They pass with **zero assertions executed**, on every CI run — on the exact axis that produced #641 and #642.

**3. One missing word in a regex.**
`compose-errors.ts:66`'s `COMPOSE_PROGRESS_RE` lists `Recreated` but not `Recreate`. `update` recreates existing containers, so Compose emits `Recreate` *first*; it fails the noise filter and masks the real error. `compose-errors.test.ts` contains **zero** occurrences of the string `Recreat` — every progress fixture uses create-path vocabulary (`Creating`/`Created`/`Starting`/`Started`). That is #644, in an otherwise excellent 23-test regression suite that is explicitly *about* message quality.

## The two lanes that are missing

Tracked separately so they can be built independently:

- **#652 — second instance on one host.** Would catch #643, #650.
- **#653 — cross-uid, actually executing in CI.** Would catch #641, #642, and un-neuters the three inert tests.

Everything else is surgical and belongs on its own issue: one word in a regex (#644), one fixture with a large file (#648), one `since`-gated migration that translates instead of deletes (#645), one carve-out in the secret audit (#649).

## Related

The architectural properties that concentrate the risk are tracked in #654, #655, #656, #657, #658. Much of this was already diagnosed in `docs/reviews/op-home-structure-issues-and-lessons.md` and `docs/operations/upgrade-hardening-plan.md` — the gap is that known items never became gates.

## Full set

Test lanes: #652 (multi-instance), #653 (cross-uid)
Architecture: #654 (migration duality), #655 (docker boundary), #656 (tree manifest), #657 (replay/backup), #658 (port collisions)
Process: #659 (release runs fewer tests than PR)
Escapes: #641 #642 #643 #644 #645 #648 #649 #650

### LINKED ISSUE #652: Missing test lane: two instances on one host (no test anywhere models a sibling instance)
Part of #651. Would have caught #643 and #650.

## Gap

Nothing in the suite ever asserts OpenPalm behaves correctly when **another instance is already running on the same host**. Searching every `.test.ts` / `.vitest.ts` / `.sh` for "second instance", "two instances", "sibling instance", "concurrent instance" returns exactly one hit — `packages/electron/test/single-instance-lock.test.ts`, which is the Electron app preventing two copies of *itself*, unrelated to two OpenPalm stacks.

Every harness that does boot something (`rootless-ownership-smoke.sh`, `rootless-host-swap-smoke.sh`, `upgrade-path-smoke.sh`, `packages/ui/e2e/global-setup.ts`) stands up exactly one `SMOKE_HOME`/`OP_HOME`.

This matters because the host is a genuinely shared resource for ports, volume names, project names, and networks — and the code is inconsistent about modelling that (see #658: `project-rename.ts` handles it carefully for project names, the port migrations don't handle it at all).

## What the lane should do

A `scripts/multi-instance-smoke.sh` that:

1. Seeds **two** isolated homes with distinct `OP_PROJECT_NAME`s, reusing the existing `smoke_write_stack_env` / `rootless-smoke-fixture.sh` helpers.
2. Brings up instance A and leaves it live.
3. Runs `openpalm update` on instance B, seeded at a prior release.
4. Asserts B does not take a port A holds, does not adopt A's project name, and — critically — that an operator override written into B's `state/stack.env` between attempts **survives** a retry.

Step 4 is the one that catches #643's second half (the override being reverted on every run).

## Also worth asserting in the same lane

`openpalm.sh` invoked from B's home acts on B and not on a project named `openpalm` (#650). The current failure mode is that it builds a *duplicate parallel stack* and dies on a port conflict, leaving an orphan `Created` container and an `openpalm_assistant-artifacts` volume behind.

## Cost

Low — the fixture helpers already exist and are already used by three smoke scripts. The new part is running two of them concurrently and asserting on the interaction, which no current harness attempts.

### LINKED ISSUE #653: Missing test lane: cross-uid (root-written files vs unprivileged CLI) — and 3 existing tests silently no-op in CI
Part of #651. Would have caught #641 and #642, and would make three currently-inert tests actually run.

## Gap

Zero `.test.ts` / `.vitest.ts` files mention "root-owned" as a concept. The one file that attempts real uid-mixing is explicit about its own limitation in its docblock:

> The mixed-axis scenarios need a root-owned axis from the resolver, so they run only under a root test process… and no-op elsewhere.

`packages/lib/src/control-plane/config-persistence-operator-ids.test.ts` guards three tests with `if (!isRootProcess()) return;` (lines 53, 70, 95). `ci.yml` runs `bun run test` on plain `ubuntu-latest` as an unprivileged user — `sudo` appears in that workflow only around disk cleanup and `docker image prune`, never around the test step.

**Those three tests therefore pass with zero assertions executed, on every CI run, in the lane that gates merges.** A test that cannot fail where it matters is worse than no test: it reads as coverage.

Meanwhile the five test files that do mention `EACCES`/`EPERM` all inject errors in-process via mocked errno objects (e.g. `akm-user-env.test.ts` spying on `chmodSync`) — none produce a real permission failure from a real file owned by a different uid.

## Why this axis specifically

The privilege boundary is real and load-bearing: containers write into bind-mounted trees as root, the CLI then has to read, copy, and delete those trees unprivileged. `core-assets.ts:196` documents it directly — *"OP_HOME/system/assistant is bind-mounted as OPENCODE_CONFIG_DIR, so the container writes into it: node_modules/…"*. Both #641 and #642 are that boundary being crossed with no net.

## Two ways to build it

**(a) Make the existing gated tests execute.** Wrap that file's run in a `docker run --rm -u root` step in `ci.yml` so `isRootProcess()` is true where it matters. Smallest change, immediately un-neuters three tests.

**(b) Drop the root requirement.** Add a fixture to `core-assets.test.ts` that `chownSync`s a file under `system/assistant/node_modules` to a synthetic uid differing from the test process, then asserts `overwriteSystemTree` surfaces a **named, actionable** error rather than a raw `EACCES` escaping to the operator. Works unprivileged in most CI containers.

(b) is preferable long-term because it makes the condition expressible in the default lane rather than a special one.

## Assert on the message, not just the throw

Both #641 and #642 surfaced as bare `EACCES: permission denied, rm '…'` / `EACCES: permission denied, copyfile '…'` with no indication of cause or remedy. The lane should assert the error names the offending path and points at `repair-ownership`. `backup.test.ts`'s existing failure-path test (*"leaves no torn final dir… when a copy fails partway through"*) proves atomicity via an injected generic throw but never asserts anything about what the operator is told.

### LINKED ISSUE #654: Migration duality: unversioned "sweep" functions run outside the versioned MIGRATIONS registry, and the lossy akm strip lives there
Root cause shared by #645 and (partly) #643. Part of #651.

## Two mechanisms, and the lossy one is unversioned

`home-schema.ts::MIGRATIONS` (lines 338-399) is a real registry: `{ since: number, run: (homeDir) => boolean }[]`, gated by `readHomeSchemaVersion`, each entry runs once. That is the right shape.

But `lifecycle.ts::applyHomeAssets` (line 183) calls a **second, structurally different** set of functions on every install/update with no version gate at all:

```ts
await applyHomeSeed(state.homeDir);
stripRetiredAkmConfigKeys(state);   // lifecycle.ts:187
reconcileDuplicateBundles(state);
ensureSystemBundle(state);
```

and `reconcileCore` unconditionally calls `pruneRemovedAddonState` (`lifecycle.ts:251`, `addons.ts:390`) on every apply — start, restart, install, update alike.

## Why that shape produces data loss

`stripRetiredAkmConfigKeys` → `stripRetiredKeysAt` → `stripRetiredAkmKeys` (`setup.ts:238`) is **delete-only**:

```ts
export function stripRetiredAkmKeys(config: Record<string, unknown>): void {
  for (const key of RETIRED_AKM_CONFIG_KEYS) delete config[key];  // includes 'profiles'
```

`RETIRED_AKM_CONFIG_KEYS` (`setup.ts:226`) includes `profiles`, `llm`, `agent`. No code path anywhere reads `profiles.llm.*` and writes the equivalent `engines.*` entry that akm 0.9 requires. The function's own docblock is candid: *"Deliberately narrow: it removes retired keys… It does not reshape bundles or defaults."*

That is the whole of #645. The instance comes up with **no LLM engine** and `openpalm validate` reports `Configuration OK.` because `validateProposedState` only checks that `state/stack.env` exists and `OP_UI_LOGIN_PASSWORD` is non-empty — it never opens `config/akm/config.json`.

## The structural point

An unversioned sweep only ever asks *"what do I delete?"*. A `since`-gated migration slot forces the author to ask *"what do I do with the old shape?"* — which is the entire difference between `stripRetiredAkmKeys` and a `migrateProfilesToEngines` that nobody wrote.

## Proposed fix

Move every "heal an upgraded install" sweep into `MIGRATIONS` with real `since` gates: `stripRetiredAkmConfigKeys`, `pruneRemovedAddonState`, `reconcileDuplicateBundles`, `ensureSystemBundle`. That buys three things at once:

1. Each gets a dedicated, individually-runnable test the way existing `home-schema.ts` entries do.
2. The slot shape makes translation the default question and deletion the explicit choice.
3. `runHomeMigrations` becomes the **one** place "did this home's shape change" is answered, so `validate` and `doctor` have a registry to cross-check instead of nothing.

Then add the missing translation itself, and extend `validateProposedState` to at least warn when `config/akm/config.json` has zero `engines` entries.

## Note on the existing test

`packages/ui/src/routes/api/assistant/akm/server.vitest.ts:425` currently **asserts the data loss is correct**. Fixing this means inverting that assertion from "the old key is gone" to "the old key survived as an `engines` entry" — worth doing in the same change so the spec and the code move together.

### LINKED ISSUE #655: No canonical docker-invocation boundary: 7 spawn paths, and the one most day-2 commands use discards stderr entirely
Root cause of #644, and of the `openpalm.sh` divergence in #650. Part of #651.

## Seven independent ways to spawn docker, with different error fidelity

`docker.ts` alone has two drivers with opposite behaviour:

**Capturing** — `run()` feeds `applyStack` (`docker.ts:1019`), which on failure calls `mapDockerError(upResult.stderr…)` (`docker.ts:1050,1099,1129`) and produces a classified, actionable message. Used by `install` / `update` via `performUpgrade`.

**Non-capturing** — `runComposeStreaming` (`docker.ts:644-664`) uses `spawn(dockerBin(), [...], { stdio: "inherit" })`. Its only error on non-zero exit is:

```ts
reject(new Error(`docker compose ${args.join(" ")} failed with exit code ${code}`));
```

Because `stdio:"inherit"` hands the fds straight to the terminal, **there is nothing in the JS process for `mapDockerError` to classify, even in principle.** Used via `activateComposeCommand` (`activation.ts:89-104`) by: `start`, `restart`, `stop`, `addon`, `rollback`, `uninstall`, `logs`.

So whether an operator gets a useful error depends on which command they happened to run.

Four further sites bypass `docker.ts` entirely with raw `execFile`/`spawn("docker", …)`, hardcoding the literal string rather than calling `dockerBin()` — which honors `OP_DOCKER_BIN`, the codebase's own documented escape hatch:

- `packages/ui/src/routes/api/setup/system-check/+server.ts:22`
- `packages/ui/src/routes/guardian/health/+server.ts:12,67`
- `packages/ui/src/lib/server/voice/bring-up.ts:209,237,278,294,304`
- `packages/cli/src/lib/host-info.ts:16,19`

An operator who sets `OP_DOCKER_BIN` is honored by the core paths and silently ignored by these four.

`packages/skeleton/openpalm.sh` is a seventh reimplementation, in shell. Its own comments are honest that reimplementing `discoverStackOverlays` in bash *"diverged from the app in ~30 ways"* and was abandoned — so it omits the conditional overlays (`guardian.compose.api.yml`, `voice.compose.lan.yml`, `workspace.compose.loopback.yml`).

## Proposed fix

1. Make `dockerBin()` plus a single captured-invocation wrapper the only sanctioned way to spawn docker anywhere in the tree. Delete the four raw call sites.
2. Give the streaming path a bounded ring buffer of recent stderr — **tee, not either/or** — so a non-zero exit can still call `mapDockerError` on something. This is what makes `start`/`restart`/`rollback` failures diagnosable at all.
3. Add the recreate vocabulary to `COMPOSE_PROGRESS_RE` (see #644) and audit the regex against Compose's actual verb list once, rather than word by word.

Item 2 removes the entire "which of seven places did this fail in" class rather than today's instance of it.

### LINKED ISSUE #656: No shared OP_HOME tree manifest: backup, purge and repair each hand-maintain their own scope list, and they have already drifted
Root cause of #648. Part of #651. Already named in your own docs as a deferred item.

## Three hand-maintained lists over one layout

`home.ts` is a genuine single source of truth for OP_HOME *paths* (its docblock: *"kill the blast-radius class of change, e.g. `stack.env` duplicated 18×"*). But **scope** over that layout — which trees matter for a given operation — is re-enumerated independently three times, in three different shapes:

| Operation | Location | Shape |
|---|---|---|
| Backup | `backup.ts:14` `UNBACKED_TOP_LEVEL = new Set(["data","cache"])` | denylist |
| Purge | `uninstall.ts` | allowlist |
| Ownership repair | `ownership-reconcile.ts:34-54` `ownershipRepairPaths` | hardcoded path list + dynamic bind-mount discovery |

They have already drifted:

- **Backup** doesn't exclude `workspace/`, so `workspace/.git` (687 MB in the reported case) is copied in full on every safety snapshot. The function's own docblock (`backup.ts:242`) claims it *"captures the user/config/state trees (config/, system/, state/, knowledge/)"* — `workspace/` isn't even mentioned in the doc, yet the code includes it.
- **Purge** already missed a tree once. Your docs record it (G7): *"`--purge` reported 'all data removed' while leaving every live credential on disk."*
- **Repair** targets the *live* `system/assistant` path and never the transient `.system-previous-<nonce>` name that only exists during `overwriteSystemTree` — which is why #641 isn't covered by `repair-ownership`.

## Already scoped, not built

`docs/reviews/op-home-restructure-proposal.md` lists *"Deferred: shared tree constant"* under "Accepted as-is", and lesson 24 states the consequence directly:

> Adding a top-level tree is an N-place change with no completeness check.

That is exactly what happened — `workspace/` was added and the backup denylist was never revisited.

## Proposed fix

One `OP_HOME_TREES` manifest declaring per tree: name, owner (cli / container / operator), durability (durable / regenerable / derived), and booleans for `inBackup`, `inPurge`, `inRepair`. `backup.ts`, `uninstall.ts`, and `ownership-reconcile.ts` all read from it.

Plus one test asserting **every tree `ensureHomeDirs` creates has an entry** — that's the completeness check lesson 24 says is missing, and it makes adding a tree a one-place change that fails loudly if you skip the declaration.

With that manifest, `workspace/` would be declared `regenerable` + `inBackup: false` and #648's 4.2 GB never happens.

### LINKED ISSUE #657: Recovery rewinds state/schema-version, so failed retries replay migrations non-idempotently and re-snapshot the full home each time
Contributes to #648 (a full snapshot per retry) and to #643's second half (operator override reverted on every run). Part of #651.

## Backups are unconditional per attempt

`backupOpenPalmHome` is called at the top of `applyManagedFiles` (`lifecycle.ts:282`), which runs fresh on **every** install/update invocation — including every manual retry after a failure. There is no hash or mtime check against the previous snapshot, and `copyEntry` does no hardlinking or content addressing, so every retry is a brand-new full byte copy of everything `UNBACKED_TOP_LEVEL` leaves in scope (see #656).

Ten attempts in eight minutes therefore produced ten snapshots. Two of them were full 1.9 GB copies containing byte-identical duplicates of the same 687 MB `.git` — verified non-hardlinked: the same relative file had different inodes across the two snapshots.

Compounding it: plain timestamp snapshots are pruned to 3 (`pruneBackupDirs(homeDir, 3)`), but `-pre-rollback` snapshots — written on **every** automatic rollback after a failed attempt (`rollback.ts:170`) — are, in the CLI's own words (`packages/cli/src/commands/backups.ts:26`), *"never pruned by anything."* That is a defensible choice for a rare event and a bad one for a retry loop.

## Recovery rewinds the version stamp

`runWithSnapshotRollback` (`lifecycle.ts:323-368`) restores a snapshot on a caught failure, and that snapshot includes `state/schema-version` (your own upgrade runbook §9 confirms the restore set). So a failed-then-recovered attempt rewinds the counter `runHomeMigrations` gates on, and the next retry replays the whole migration chain.

For migrations whose condition is **content-based rather than idempotency-flagged**, replay is not a no-op. `migrateConsolidatedDefaultPorts` (`config-persistence.ts:173-199`) fires on `assistantPort === "3800" && (uiPort === "3810" || !uiPort)` — a live-file check. An operator who fixes the port by hand between attempts has it silently reverted on the next run, which is precisely the reported behaviour in #643.

## Proposed fix

1. **Hash-gate the backup.** Skip the copy when a content hash of the in-scope tree matches the newest existing snapshot. Turns "one full copy per retry" into "one copy per actual change" and fixes the storm without changing retention policy.
2. **Prune `-pre-rollback` too**, or at least cap it — the "never pruned" rule was written for a rare event.
3. **Decouple rollback-restore from migration-replay.** Either stop bundling `state/schema-version` into the restore set, or have migrations re-diff live state against their preconditions rather than relying solely on `since >=`. The second is more robust: it makes replay safe by construction rather than by careful bookkeeping.

### LINKED ISSUE #658: Port migrations are host-blind: the collision-safe pattern used for project names is never applied to ports, though the prober already exists
Root cause of #643's first half. Part of #651. **Cheapest fix of the architectural set — the working pattern already exists in-tree.**

## Multi-instance safety is solved for project names and not attempted for ports

`project-rename.ts` is careful work: it records rename intent, and before acting it verifies via the Compose `working_dir` label that a same-named project actually belongs to *this* install (`teardownRenamedProject`, lines 100-193). `deploy.ts:227` separately refuses to deploy over a same-project-name stack rooted at a different `OP_HOME`. The codebase clearly knows how to make a host-shared resource collision-safe.

Ports get none of it. `STACK_DEFAULTS.ports` (`defaults.ts:7-16`) is one fixed constant — `ui:3800, assistant:3810, workspace:3820, hostUi:3880` — so every OP_HOME on a machine that doesn't manually override lands on the same four ports.

A real, host-aware prober **already exists**: `checkPortAvailable` (raw TCP bind) and `portHeldByOurContainer` (folds in Docker ownership so your own stack doesn't false-positive) in `port-probe.ts:24-96`. It is wired into `openpalm doctor` and the install wizard (`resolveInstallPortTargets`, `port-probe.ts:125-132`).

It is **not** wired into the migrations that reassign those same ports during an upgrade. `migrateLegacyDefaultPorts` (`config-persistence.ts:80-110`) and `migrateConsolidatedDefaultPorts` (lines 173-199) write `STACK_DEFAULTS.ports.*` unconditionally once their string-match condition holds, with zero knowledge of what else is listening. Deterministic per instance, blind to the host.

## What that looked like in practice

Three instances on one host. The upgrade moved instance B's assistant onto 3810, which instance C's UI already held. `docker compose up` then failed with a raw `Bind for 127.0.0.1:3810 failed: port is already allocated` — surfaced through the driver that discards stderr (#ARCH_DOCKER), so the operator saw only `Container B-assistant-1 Recreate` (#644).

`openpalm doctor` *did* report `assistant (3810): CONFLICT` — but it reports its own instance's live ports the same way, so a genuine cross-instance collision is indistinguishable from normal operation. Worth fixing alongside: distinguish "in use by this instance" from "in use by another process".

## Proposed fix

Route both port migrations through `probeInstallPorts` before writing. If the default collides, pick and persist the next free port instead of writing the fixed default and letting `up` fail later.

This is a small change to two functions, reusing a prober that already exists and is already trusted by two other call sites.

### LINKED ISSUE #659: release.yml runs fewer tests than ci.yml: the Docker smokes, upgrade-path-smoke and Tier 4 are all PR-only
Part of #651.

## The release pipeline runs a narrower set of tests than an ordinary PR

Reading `.github/workflows/ci.yml` and `release.yml` side by side:

**`ci.yml` `quality-gates` (runs on PR)** additionally executes several Docker smokes that are outside the `bun test` tier system entirely:

- `validate-rootless-guardrails.sh`
- `rootless-ownership-smoke.sh` (two targets)
- `rootless-host-swap-smoke.sh`
- `guardian-image-offline-smoke.sh`
- **`upgrade-path-smoke.sh`** — the closest thing in the repo to a real prior-release upgrade test

**`release.yml` `preflight`** runs only: `bun run test`, `ui:check`, `ui:test:unit`, `electron:test`.

Grepping `release.yml` for `upgrade-path-smoke`, `RUN_DOCKER_STACK_TESTS`, `dev-e2e-test`, `playwright` returns **zero matches**. So does Tier 4 (`ui:test:e2e:mocked`), which `ci.yml` gained recently — its own comment records why:

> Tier 4 was documented in the release runbook but run by nobody: CI had no Playwright lane for it… a PR breaking a browser-level path merged green

Tier 5 (live Docker stack, 8 `*.stack.ts` specs) is gated behind `RUN_DOCKER_STACK_TESTS=1` and runs in neither workflow. `ci.yml` says so directly: *"Tier 5 (live stack) still is not covered; it needs a daemon and images."*

## Why it matters

This presumably relies on branch protection having run `ci.yml` before the release branch is cut. That holds until it doesn't — a hotfix branch, a re-tag, a manual dispatch, or a protection rule change all bypass it silently. The pipeline that produces the artifact users install should be the strictest one, not the most permissive.

## Also worth fixing while here

`upgrade-path-smoke.sh` is the right idea but incomplete by your own assessment. `docs/operations/upgrade-hardening-plan.md` (Mode 3) records that it *"never calls akm… never runs `applyHomeAssets` (only the seed)… force-deletes `state/schema-version`"*, which makes the mid-cycle stamp population untestable by construction — and that stamp is exactly what #657 shows replaying migrations non-idempotently.

## Proposed fix

1. Add the Docker smokes and Tier 4 to `release.yml`'s `preflight`, or have `preflight` invoke the same job `ci.yml` uses so the two cannot drift.
2. Close `upgrade-path-smoke.sh`'s three documented gaps so "we tested the upgrade path" means the akm binary and `applyHomeAssets` actually ran.
3. Decide explicitly whether Tier 5 gates a release. If it does, it needs a runner with a daemon; if it doesn't, the runbook should stop implying it does.

itlackey:
## Summary

Follow-up to #647, covering the full 0.13.1 milestone: the five band-aids from #647 are reverted and replaced with design fixes, and the nine systemic issues (#651–#659) plus the remaining escapes (#648, #649, #650) are addressed. 44 commits, 98 files, +5923/−2202.

Closes #639, closes #641, closes #642, closes #643, closes #644, closes #645, closes #648, closes #649, closes #650, closes #651, closes #652, closes #653, closes #654, closes #655, closes #656, closes #657, closes #658, closes #659, closes #632, closes #633. (Closing keywords do not fire on a non-default base; each issue has a comment with evidence.)

### Band-aids reverted, design fixes landed

| Issue | Was | Now |
|---|---|---|
| #639 | `openpalm unpin`, clear endpoint, button | Removed. The `rollback-` pin already releases on the next successful update; the test proves it. Read-only notice stays. |
| #641 / #642 | Best-effort `rmSync`; tolerant backup walker | Ownership repair covers `system/` and staging leftovers, its marker fingerprints the path set, install/update force the walk. Swap and backup are strict again, and an EACCES names the path and `openpalm repair-ownership`. |
| #632 | Audit existence check | `--no-start` without `--file` rejected before any disk write. |
| #633 | Regex patch of `bun.lock` | Removed; bun neither maintains nor reads that field. |

### Systemic (#651)

- **#654** heals are `since`-gated migrations (schema 12); the UI test that specified the data loss now asserts the translation; a replay test proves re-running the chain after a rollback never touches an operator edit (#657.3).
- **#658 / #643** the port migrations write a default only after `probeInstallPorts` says it is free or ours; explicit values are never moved; `doctor` names "held by another process". The new multi-instance lane then caught the rest of the class on its first run (the shared workspace default), so `ensureHostPortDefaults` now runs on every install/update for every compose-published port key with no explicit value (`HOST_PORT_DEFAULTS` is the one list), and a fresh `stack.env` no longer bakes the ui/assistant/workspace defaults as if an operator had chosen them.
- **#656 / #648 / #657** `OP_HOME_TREES` manifest drives backup, purge, and repair scope with a completeness test; `workspace/` is out of the backup; unchanged homes are not re-snapshotted (content hash); `-pre-rollback` snapshots capped; `openpalm backups list` exists.
- **#655 / #644** `runComposeStreaming` tees stderr and classifies through `mapDockerError`; the four raw `docker` spawns go through `dockerBin()` with a grep guard; the progress filter audited against a real Compose v5 daemon; port conflicts classified whichever side the port is printed on.
- **#650** `openpalm.sh`/`.ps1` rendered by the control plane from the resolved overlay list and project name; refuses under the wrong `OP_HOME` or with no project.
- **#652 / #653 / #659** `multi-instance-smoke.sh` and `cross-uid-smoke.sh` in Quality gates; the three root-gated tests execute under sudo and fail loudly if skipped; `gates.yml` (`workflow_call`) shared by `ci.yml` and release `preflight`; `upgrade-path-smoke.sh` runs `applyHomeAssets` and the pinned akm and keeps the historical schema stamp; Tier 5 explicitly does not gate a release.
- **#649** audit messages name `knowledge/env/user.env` for env-consumed credentials.
- Also: `performUpgrade` gets `runDeploy`'s dev-tag pull policy (the multi-instance lane exposed it; fixed in source, not shimmed).

### Verification

- CI on the head is green across all six gates. Quality gates ran the new lanes for real: the multi-instance lane updated instance B beside a live instance A and asserted ports, project name, workspace port, and a mid-retry operator override; the cross-uid lane repaired root-owned files with the real sandboxed chown and exercised the strict swap and backup as the unprivileged runner; the upgrade-path smoke ran `applyHomeAssets` and the pinned akm binary against 0.12.43 and 0.13.0-beta.13 homes.
- Local, on the head: `bun run lint` clean; `bun run test` 2681 pass / 0 fail; `bun run ui:check` 0 errors; electron 188 pass; permission-sensitive suites also run as an unprivileged user.
- Every change has before/after evidence on a real isolated `OP_HOME` through the real CLI/lib entry points; see the per-issue comments.
- **Branch protection:** reusable-workflow checks report as `gates / Quality gates`, `preflight / Biome lint`, etc. Rules referencing the old bare names need updating.

https://claude.ai/code/session_018rQ2BpDPUCde6Na4gZdKpu


itlackey:
**CI note: `gates / Quality gates` is red on 869313f4 because the new multi-instance lane (#652) caught a real defect, not a script problem.**

Instance B's real `openpalm update` failed with `Port 3820 is already in use by another program. Free it, then retry.` (classified correctly by the #655 work). 3820 is the workspace default; neither instance has `OP_WORKSPACE_PORT` in `state/stack.env`, so both fall back to compose's `${OP_WORKSPACE_PORT:-3820}` and collide. The port migrations only ever handled assistant/ui, and a fresh install writes the workspace default blindly, so this is the #658 host-blind class on a port nothing considered. Everything else the lane asserts passed: the operator override survived the retry, B kept its project name and its five seeded ports, A stayed healthy, and `openpalm.sh` from B's home resolved B's project.

Fix in progress on this branch: one host-aware step on install/update that, for every compose-published port key with no explicit value, probes the default through the existing prober and persists the next free port only on collision (explicit values stay authoritative), plus a lane assertion on the workspace port. The earlier red on 833d43a8 was the cross-uid smoke's spec file missing a `.yaml` suffix, fixed in b8335ec8; that lane now runs past install.

---

itlackey:
Green on aa1b33cc. Closing the loop on the CI note above: the multi-instance lane's finding is fixed in source (64cb95d5 `ensureHostPortDefaults` for every published port key on install/update; fde9937b no baked port rows in a fresh `stack.env`, with a CLI-path test), and on this run instance B's real update completed and came up healthy beside live A, with every assertion passing. The cross-uid lane also ran for the first time with the real sandboxed chown and passed. Two script-only fixes along the way: the spec file suffix and two `set -e`/`pipefail` aborts that were hiding the lanes' own diagnostics.

---

itlackey:
## akm-cli 0.9.7 → 0.9.8-beta.2 (1e10b379) — bump gate evidence

**Pins moved in lockstep:** `containers/assistant/tools/package.json`, `packages/skeleton/system/paperclip/package.json` (`akm-opencode` unchanged at 0.9.2202608290901). Engines `node >=22` vs image base `node:22-trixie-slim`: OK. Unchanged contracts re-verified on the installed dist: `configVersion "0.9.0"` only, `018-drop-dead-lane-schema` still `historical-destructive`, `akm migrate status` exit/`status` contract, and the literal ``Run `akm upgrade --force` `` the entrypoint greps.

**`openpalm-akm-state-upgrade` is now a wrapper over the supported surface.** akm#895 shipped `akm upgrade --state-only` in 0.9.8; the helper execs the image-pinned `akm --format json upgrade --state-only` instead of importing `dist/core/state-db.js`. Test asserts the single exec line, the state-only form, no `--force`, no dist internals. Boot behavior unchanged (report + point, never apply).

### CI on 1e10b379 (all six `gates / …` green)
`gates / Quality gates` ran the real pin under the image's node:
```
=== Pin under test: akm-cli 0.9.8-beta.2 on node:22-trixie-slim ===
PASS: akm-cli@0.9.8-beta.2 installs and runs under node:22-trixie-slim (0.9.8-beta.2)
PASS: migrate status on the shipped home is "current" (must not be "blocked")
PASS: the shipped home reconciles at least one task on its own (got 3: akm-improve, prompt-assistant, session-maintenance)
PASS: the unconvertible fixture is reported as a failure (failures: assistant-daily-briefing.yml, zz-unconvertible-fixture.yml)
PASS: adding one unconvertible file breaks no other task (regressed: none)
PASS: and introduces no new failures beyond itself (new: none)
akm pin integration smoke passed for akm-cli@0.9.8-beta.2 on node:22-trixie-slim.
ok: 0.12.43: real akm migrate status is 'current' (not blocked) against the migrated home
ok: 0.12.43: real akm task sync reconciled 3 shipped task(s) from the migrated home
ok: 0.13.0-beta.13: real akm migrate status is 'current' (not blocked) against the migrated home
ok: 0.13.0-beta.13: real akm task sync reconciled 3 shipped task(s) from the migrated home
```
(`assistant-daily-briefing.yml` fails in both phases for "No engine was selected and the fixed opencode-sdk fallback is unavailable" — the CI runner has no engine; the smoke is differential so it cancels out. Same as before the bump.)

### Host-side, real binaries (node v22.22.2), scratch homes only
| Scenario | Result |
|---|---|
| Pin-smoke replica of `akm-pin-integration-smoke.sh` on 0.9.8-beta.2 | 9/9 assertions PASS |
| 0.9.7 creates state.db → 0.9.8-beta.2 `migrate status` current, `health` opens it (exit 4: new `data-dir-usage` advisory), `task sync` reconciles the same 3 tasks, `upgrade --state-only` → "already current" | PASS |
| 0.9.7 reopens that state.db | exit 78: `Refusing to open a database with a newer migration ledger: unknown migration ID 025-task-history-vocabulary-backfill` — **one-way confirmed**; documented with a `VACUUM INTO` snapshot step in `docs/managing-openpalm.md` because `data/` is not in a backup |
| Historical-destructive cutover (pre-018 state.db from akm 0.8.14; 0.9.x cannot produce one, a fresh db is created at the current ledger) → 0.9.8-beta.2 `health` exit 78 naming 018 and both remedies → `akm upgrade --state-only` | `stateUpgrade.applied: true`, `.bak` sibling present, `PRAGMA quick_check` ok, `health` then exit 0, second run "already current" — PASS |
| `task sync --rebind` npm probe (akm#901), counted via `strace` execve | 0.9.7: 2 spawns of `npm root --global`; 0.9.8-beta.2: 1 — fixed |
| UI engine shape with `extraParams: {"top_p": 0.9}` + first-class temperature/maxTokens | `config list` / `health` accept it — PASS |

### Two findings to carry upstream (not fixed here)
1. **Circular refusal on a legacy `extraParams` key.** With `engines.<x>.extraParams.temperature` present, 0.9.8-beta.2 exits 78 from `config list`, `migrate status` **and `migrate apply`** with "Run `akm migrate apply` to rewrite the config file" — the named remedy is blocked by the same load-time check (`dist/core/config/config.js:166`). The only way out is a hand edit. OpenPalm's own writers never emit those keys, but the Automations/AKM settings `extraParams` box is free-form JSON, so an operator who typed `temperature`/`max_tokens`/`enable_thinking`/`reasoning_effort` there would have every akm command (and the boot check) fail after this bump.
2. **Blocked v2 task text** (#902/#899 claims to name the `command:` → `run:` + `shell:` rewrite): what 0.9.8-beta.2 prints is the reason slug `shell-command-resolution-changes-v2-literal-argv-semantics` plus "needs a human decision … review the file"; the field rewrite is not spelled out.

### Prerelease caveat (in CHANGELOG)
`akm-opencode` selects its CLI by probing `akm --version` against `^0.9.2`; node-semver never admits a prerelease, so on this beta the plugin falls back to the akm-cli 0.9.7 it bundles — which then refuses a state.db 0.9.8 has opened. The stable 0.9.8 satisfies the range. The pin must move to `0.9.8` before the release tag; the CLI-side verification above was done on the beta.

---

itlackey:
## Pin moved to akm-cli 0.9.8-beta.3 (44a332cf)

Supersedes the beta.2 evidence above. beta.3 is fix-only, and two of its three fixes are the defects the bump verification reported upstream.

**Fixed upstream, re-verified here against the real binary on node 22:**
| Finding (reported on beta.2) | beta.2 | beta.3 |
|---|---|---|
| Legacy `extraParams` key made every command exit 78, including the `akm migrate apply` the error names | `migrate status` / `migrate apply` / `config list` all exit 78 | `migrate status` exits 1 with the blocker `engines.default.extraParams.temperature -> engines.default.temperature`; `migrate apply` exits 0 and rewrites the key onto the first-class field; `config list` then exits 0 |
| `data-dir-usage` warned on an untouched install (`-wal`/`-shm` counted against the data dir but not their own databases) | `health` exit 4 on a fresh install | `health` exit 0, status `pass` |

Third fix, not one of ours: the incremental index's per-directory freshness digest now covers every file over `(basename, size, mtime, ctime)`, so an edit with a restored mtime can no longer leave stale content indexed. Costs one extra rescan after a `chmod` or a tree copy, and one as existing indexes take the new digest shape.

**Unchanged contracts** (checked against the beta.3 dist, not assumed): state migration ledger byte-identical to beta.2 — no new migration, so the one-way state.db caveat and its `VACUUM INTO` snapshot step stand as written; `018-drop-dead-lane-schema` still `historical-destructive`; `configVersion "0.9.0"` still the only accepted contract; `akm upgrade --state-only` present; the health refusal still carries the literal ``Run `akm upgrade --force` `` the entrypoint greps.

**CI on 44a332cf — six of six green.** Quality gates ran the real pin:
```
=== Pin under test: akm-cli 0.9.8-beta.3 on node:22-trixie-slim ===
PASS: akm-cli@0.9.8-beta.3 installs and runs under node:22-trixie-slim (0.9.8-beta.3)
akm pin integration smoke passed for akm-cli@0.9.8-beta.3 on node:22-trixie-slim.
ok: 0.12.43: real akm migrate status is 'current' (not blocked) against the migrated home
ok: 0.12.43: real akm task sync reconciled 3 shipped task(s) from the migrated home
ok: 0.13.0-beta.13: real akm migrate status is 'current' (not blocked) against the migrated home
ok: 0.13.0-beta.13: real akm task sync reconciled 3 shipped task(s) from the migrated home
```
Local on the head: lint clean, 2248 lib/cli/guardian tests pass, `ui:check` 0 errors, scripts 84 pass, plus a host replica of the pin smoke against a materialized `OP_HOME` scoring 9/9 on the script's own assertion block.

**Still open upstream (cosmetic):** a blocked v2 task's `migrate status` blocker prints only the reason slug `shell-command-resolution-changes-v2-literal-argv-semantics`, not the `command:` argv → `run:` + `shell:` rewrite the beta.2 notes describe. Nothing in OpenPalm depends on that wording.

**Still blocking the tag:** the akm-opencode plugin's `^0.9.2` range admits no prerelease, so the pin must move to the stable 0.9.8. The UI-side `extraParams` lift (15062ae7) stays regardless — akm still fails config load closed on those keys, so the settings tab must not write one.

---

itlackey:
## Final pin: akm-cli 0.9.8 + akm-opencode 0.9.8202609020939 (840e9717)

Supersedes the beta evidence above. Both halves of the akm lane are released, so every lockstep surface moves together and **the prerelease caveat this branch carried is gone**.

| Surface | Was | Now |
|---|---|---|
| `containers/assistant/tools/package.json` | `akm-cli` 0.9.8-beta.3 | **0.9.8** |
| `packages/skeleton/system/paperclip/package.json` | `akm-cli` 0.9.8-beta.3, `akm-opencode` 0.9.2202608290901 | **0.9.8**, **0.9.8202609020939** |
| `packages/skeleton/system/assistant/opencode.jsonc` | `akm-opencode@0.9.2202608290901` | **`@0.9.8202609020939`** |
| `paperclip-addon-design.md`, `manual-tailscale-paperclip-testing.md`, `CHANGELOG.md` | beta pin + caveat | stable pin, caveat deleted |

**Why the plugin had to move in the same commit.** akm-opencode picks its CLI by probing `akm --version` against a compatibility range. The old plugin's range was `^0.9.2`, which node-semver never lets a prerelease satisfy, so on a beta pin it skipped the image's akm and fell back to the akm-cli 0.9.7 it bundles — which then refuses a state.db 0.9.8 has opened. The new plugin declares `AKM_VERSION_RANGE = "^0.9.8"` (dependency `akm-cli: ^0.9.8`), verified here to accept 0.9.8 and reject 0.9.7. It now selects the image's own akm. That is the whole reason this bump waited for the stable release.

**The CLI side needs no re-verification, and that is a fact rather than an assumption:** `akm-cli@0.9.8`'s `dist` is **byte-identical to 0.9.8-beta.3** — `diff -rq` over the whole tree returns nothing. The published CHANGELOG folds the three beta entries into one release note and adds no code. So every result from the beta.3 evidence above carries over exactly. Re-checked anyway against the released tarball: state migration ledger unchanged (still 025 and 026, no new migration, so the one-way state.db caveat and its `VACUUM INTO` step stand as documented); `018-drop-dead-lane-schema` still `historical-destructive`; `configVersion "0.9.0"` still the only accepted contract; `akm upgrade --state-only` present; the health refusal still carries the literal ``Run `akm upgrade --force` `` the entrypoint greps.

**CI on 840e9717 — six of six green.** Quality gates ran the real pin:
```
=== Pin under test: akm-cli 0.9.8 on node:22-trixie-slim ===
PASS: akm-cli@0.9.8 installs and runs under node:22-trixie-slim (0.9.8)
PASS: migrate status on the shipped home is "current" (must not be "blocked")
PASS: the shipped home reconciles at least one task on its own (got 3: akm-improve, prompt-assistant, session-maintenance)
PASS: the unconvertible fixture is reported as a failure (failures: assistant-daily-briefing.yml, zz-unconvertible-fixture.yml)
akm pin integration smoke passed for akm-cli@0.9.8 on node:22-trixie-slim.
ok: 0.12.43: real akm migrate status is 'current' (not blocked) against the migrated home
ok: 0.13.0-beta.13: real akm task sync reconciled 3 shipped task(s) from the migrated home
```
Local on the head: lint clean; 2248 lib/cli/guardian tests pass, including `paperclip-compose-contract` which cross-checks the Paperclip manifest against `opencode.jsonc`; `ui:check` 0 errors; scripts 84 pass; host replica of the pin smoke against the real 0.9.8 binary 9/9, fresh-install `akm health` exit 0.

The two defects this repo's verification reported against the betas ship fixed in the release: `akm migrate apply` can now clear a legacy `extraParams` config, and `data-dir-usage` no longer warns on an untouched install. The UI-side `extraParams` lift (15062ae7) stays regardless — akm still fails config load closed on those keys, so the settings tab must not write one.

---

